PR #12, opened 2017-09-04
Maximize cache hits

Kikobeats:
Hello,

I noted you are using an LRU cache, that's great, but with more than 350K packages at npm registry, keeping a cache with a max of 500 is too small.

For maximize hits, I added [lru-ram](https://github.com/Kikobeats/lru-ram) for calculate the optimal number of items to keep into the cache, based on the machine RAM and a fixture of data that represent a single item.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "koa-router": "^7.1.0",
     "koa-static": "^3.0.0",
     "lru-cache": "^4.0.2",
+    "lru-ram": "~1.1.0",
     "normalize.css": "^5.0.0",
     "performance-now": "^2.1.0",
     "preact": "^7.1.0",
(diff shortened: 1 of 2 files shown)


pastelsky:
Thanks for the PR. On a server with limited RAM, the free RAM space can change dynamically based on load characteristics. We might end up reserving too much space upfront when the server is initied. Is it possible to change this dynamically?

Kikobeats:
hmm I understand your throughts but not sure if your comment is a good idea.

`lru-ram` just calculate the optime number of items based on a quantity of ram. If you want to resize the LRU you need to create/destroy all the time.

On the other hand, the OS automatically will be move ram content to filesystem when ram is 100%.

Probably the number need to be adapted based on the server used

pastelsky:
Closing this as I don't believe using a regular cache vs. lru-ram is going to make a lot of difference here. In case of a memory cache miss, we'll refer to our persistent datastore which is fast and access cheap enough. 

I prefer the keep `lru-cache` which has been battle-tested. Thanks for your PR.

PR #32, opened 2017-12-18
Fix wrong URL hyperlink

LucaT1:
Changes the redirection from `https://github.com/pastelsky/bundlephobia/yarnpkg.com` to the correct url(`https://yarnpkg.com`)


pastelsky:
Thanks. I can never seem to get links right :P

PR #79, opened 2018-08-15
Autofocus on input field!

itsfadnis:
Saves a click or tab before you get typing 🤷‍♂️


tom-sherman:
Probably want to consider how this behaves on mobile - a keyboard popping up right after the page loads is not ideal.

itsfadnis:
@tom-sherman good point 👍

tom-sherman:
@itsfadnis I think it's still a good idea though for desktop, maybe you could take a crack at implementing it desktop only?

pastelsky:
Thanks for the PR!

PR #88, opened 2018-09-05
added support for open search

### LINKED ISSUE #86: [feature request] open search description
## Feature request: Open Search Description
It would be great to search packages directly from the search bar as seen in this example from stackoverflow:

screen shot 2018-09-01 at 00 45 45

This is also easy to implement, see https://developer.mozilla.org/en-US/docs/Web/OpenSearch for more information.

Absolutely loving the site btw 👍

leoweigand:
I was not able to actually test it, but it should work fine.
Closes #86


pastelsky:
Thanks for the PR!

Issue #59, opened 2018-03-31
Error running dev server: The opts.token property is missing.


ndrewr:
## Type
Bug
## Package name
le_node

### Entire error (stringified) I see in my browser console
```
bundlephobia/node_modules/le_node/lib/logger.js:176
      throw new _error.BadOptionsError(opts, _text2.default.noToken());
      ^

Error: The opts.token property is missing.
```

I added the API key from Algolia as instructed in the readme but it seems like this logging service API key is also needed?

Locally I can just modify `index.js` (or sign up for LogEntries I suppose), i.e.
```
const log = new Logger({
  token: process.env.LOGENTRIES_TOKEN,
});
```

but should we just make this optional?

pastelsky:
LOGENTRIES has been removed from bundlephobia, this should no longer be an issue.

PR #96, opened 2018-09-21
replace le_node with console.log and console.error

ansballard:
if LOGENTRIES_TOKEN isn't specified

fixes https://github.com/pastelsky/bundlephobia/issues/59


pastelsky:
Logentries has been removed in the meanwhile. Thanks for the PR.

PR #110, opened 2018-11-02
Add more markdown parsers

wooorm:
Markdown parsers are now sorted by downloads on npm.
Tags are now sorted from high-to-low.

### DIFF
--- a/server/middlewares/similar-packages/fixtures.js
+++ b/server/middlewares/similar-packages/fixtures.js
@@ -35,14 +35,14 @@ const categories = {
   'markdown-parser': {
     name: 'Markdown parsers',
     tags: [
-      { tag: 'markdown', weight: Weight.HIGH },
-      { tag: 'parse', weight: Weight.NORMAL },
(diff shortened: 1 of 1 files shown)


pastelsky:
Thanks for your suggestion. `Markdown parser` is a crowded category and we wish to keep suggestions limited to 4-5 for them to remain effective. We do not aim to exhaustively list all libraries for a purpose. We wouldn't be able to incorporate this at the moment.

wooorm:
That makes sense, but to add: I maintain both the most popular (remark) and smallest commonmark compliant (micromark) markdown parsers, I feel it’s unfair to users to give outdated advice, which are now bigger or less popular choices. For example, you‘re not suggesting to use Moment.js either, and not suggesting luxon/dayjs/date-fns, just because there are already so many date libraries?

I totally understand that your busy and this isn‘t your main focus (I mean, I raised this two years ago and only now got a comment, that’s cool). But you have a list of suggestions for markdown parsers: folks are basing their choices on ’em. I think those suggestions should be good suggestions.

PR #117, opened 2018-11-19
Fix title of dependency.svg on hover

styfle:
Hovering over the icon was showing "Group 11" which was a bit confusing.
This will change the hover to display "Dependencies".


pastelsky:
Thanks for the fix, and sorry for the delay in merging it :)

PR #126, opened 2018-12-28
Add JSS to the alternatives

oliviertassinari:
cc @kof


pastelsky:
Thanks for the PR. Looks like `JSS` has already been added to the `CSS in JS libraries` category in the meanwhile.

PR #139, opened 2019-02-18
Fix react-pdf put in client-side PDF creation (naming confusion)

wojtekmaj:
React-PDF (react-pdf) is used to display PDFs in browser. @diegomura's React-PDF (@react-pdf/renderer) is used to create PDFs.


pastelsky:
Thanks for the fix, and apologies in the delay in merging this.

PR #146, opened 2019-03-02
Fix build badge to link to travis instead to img

ultrox:
Not sure is this on purpose but I fix it anyway. My intention was to see why it is failing and got linked to img instead. Other day I linked to img by accident, so here if you want it.


pastelsky:
Thanks!

PR #213, opened 2019-08-30
Fix multiplier calculation + minor improvements

### LINKED ISSUE #212: Incorrect size comparison for differences over 100%
**Describe the bug**
When size difference between two packages more than 100%, difference is shown as incorrect multiplier

**To Reproduce**
https://bundlephobia.com/result?p=recharts@1.7.1 (120.5kB min + gzip) comparison with `echarts` (258.5kB min + gzip) shows `1.15x larger`. However 120.5 * 1.15 is only 138.57 and not 258.5

**Expected behavior**
This is true for percentage (it is 115% bigger), but multiplier should be `2.15x larger` then

ViLPy:
Fixes #212 

- size multiplier was calculated incorrectly from percentage value;
- fix indentation;
- remove unnecessary ternary operator since comparisonSizePercent is always greater than 0 for this block;

### DIFF
--- a/client/components/SimilarPackageCard/SimilarPackageCard.js
+++ b/client/components/SimilarPackageCard/SimilarPackageCard.js
@@ -42,14 +42,14 @@ export default class SimilarPackageCard extends Component {
             </div>
           </div>
         )
-      } else if (comparisonSizePercent > 100) {
+      } else if (Math.abs(comparisonSizePercent) > 100) {
         return (
           <div>
             <div className="similar-package-card__number">
               {1 + Math.abs(comparisonSizePercent).toFixed(0) / 100} <span className="similar-package-card__shrink">×</span>
             </div>
             <div className="similar-package-card__label">
-              Larger
+              {comparisonSizePercent > 0 ? 'Larger' : 'Smaller'}
             </div>
           </div>
         )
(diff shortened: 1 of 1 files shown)


client/components/SimilarPackageCard/SimilarPackageCard.js
```diff
@@ -46,10 +46,10 @@ export default class SimilarPackageCard extends Component {
         return (
           <div>
             <div className="similar-package-card__number">
-              {Math.abs(comparisonSizePercent).toFixed(0) / 100} <span className="similar-package-card__shrink">×</span>
+              {1 + Math.abs(comparisonSizePercent).toFixed(0) / 100} <span className="similar-package-card__shrink">×</span>
             </div>
             <div className="similar-package-card__label">
-              {comparisonSizePercent > 0 ? 'Larger' : 'Smaller'}
```
pastelsky: Instead of removing this, I think the `else if` should use the absolute value of `comparisonSizePercent`. The goal is to show the difference in terms of `<times> x` when it is more than 100% in either direction.

ViLPy: Fixed

PR #229, opened 2019-10-15
Fix apostrophe escape typo in BarGraph.js

jakejarvis:
Hello! A very, very simple PR here... 

A string in BarGraph.js (for the indicator above the bars that indicates side-effect free w/o ESM) read `doesn\t` instead of `doesn\'t`, causing a tab to appear instead of an apostrophe.

Before: (from https://bundlephobia.com/result?p=@sindresorhus/is@1.2.0)

Screen Shot 2019-10-12 at 10 39 39 AM copy

After:

Screen Shot 2019-10-12 at 10 52 46 AM copy

Thanks! :)

### DIFF
--- a/client/components/BarGraph/BarGraph.js
+++ b/client/components/BarGraph/BarGraph.js
@@ -93,7 +93,7 @@ export default class BarGraph extends PureComponent {
             )}
             {options.isFirstSideEffectFree && (
               <div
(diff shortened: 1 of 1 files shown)


pastelsky:
Thanks for the PR Jake.

PR #235, opened 2019-10-27
Fix ProgressHex mount blocking render for several seconds

### LINKED ISSUE #234: Progress hex animation blocks render on mount
**Describe the bug**
The mounting sequence for the Progress Hex blocks the main thread for 1-10 seconds due to `anime` timeline add method. This especially impacts mobile users as the user interface completely freezes for 2+ seconds (on Chrome for Android, running on Google Pixel XL). 

**To Reproduce**

1. Go to home page
2. Search for a package, immediately try to use the UI before loading indicator mounts (i.e. scroll)

For debugging/testing:
1. Set profiler to record.
2. Perform the above steps.

Screen Shot 2019-10-27 at 8 59 28 AM

**Expected behavior**
Progress indicator should not block render or freeze the UI.

mvasigh:
Fixes #234 

Seems that `ProgressHexAnimator` mounts to the `anime` timeline redundantly in a loop, where a single invocation does the same thing. This brings `tl.add()` runtime down to 42ms with the same profiler settings, down from 6000ms+

Screen Shot 2019-10-27 at 9 12 06 AM

### DIFF
--- a/client/components/ProgressHex/progress-hex-timeline.js
+++ b/client/components/ProgressHex/progress-hex-timeline.js
@@ -81,9 +81,7 @@ export default class ProgressHexAnimator {
     }
 
     timeline.add(fadeInRings)
-    for (let i = 0; i <= 50; i++) {
-      timeline.add(quakeCircles)
-    }
+    timeline.add(quakeCircles)
     return timeline
   }
 }


pastelsky:
Good catch @mvasigh . I've felt the slowdown in response times as well after having added this. However, removing this loop changes the animation –

**Before** 
after


**After**
before

We'll need a different way to solve this. Maybe instead of queuing up 50 iterations upfront, its best to add the next iteration in the `onFinish` (or similar) callback of the previous iteration – one at a time.

mvasigh:
Good point! I assumed that the second portion of the animation is intended to loop (rather than just run 50 times) so I separated the two into different `anime` instantiations and return a Promise that resolves to the timeline.

Before: 
Screen Shot 2019-10-29 at 12 36 01 PM

After (showing both the first "half" and the timeline with looping animation):
Screen Shot 2019-10-29 at 12 34 46 PM
Screen Shot 2019-10-29 at 12 37 24 PM

However, since we are no longer adding multiple stages, seems like we don't need a `timeline` instance and should just loop the second portion? Should I just remove the use of `timeline`? (haven't worked with this lib before)

Yedfree (approved):

PR #241, opened 2019-11-07
Improve <BarGraph/> styling

Andarist:
**Before**
Screenshot 2019-11-07 at 17 12 08

As we may notice there are 2 problems here:
- version label pushes the bar so one might come to wrong conclusions solely by looking at bars' heights
- RTL used for layout purposes~ makes that label look weird (a hyphen is misplaced)

**After**
Screenshot 2019-11-07 at 17 26 38


pastelsky:
Sweet, thanks for the fix!

Issue #281, opened 2020-01-21
Image preview in the package README breaks the bundlephobia layout


bpierre:
See:

- https://bundlephobia.com/result?p=@aragon/ui@1.2.1
- https://github.com/aragon/aragon-ui

PS: thanks for having built this super valuable tool! :hugs:

aviskarkc10:
One fix for this is to forbid the `img` tag when sanitizing dom.

https://github.com/pastelsky/bundlephobia/blob/bundlephobia/client/components/QuickStatsBar/QuickStatsBar.js#L35
in the above line we could do this.
```
    return DOMPurify.sanitize(trimmed, { FORBID_TAGS: ['img']})
```
what do you think of this solution @pastelsky

pastelsky:
@aviskarkc10 sounds good. Do you want to take a stab?

aviskarkc10:
yeah sure.

pastelsky:
Closing as completed: the current package page does not render package README content, so the reported README-image layout path is no longer present and cannot be reproduced.

Issue #242, opened 2019-11-09
Similar packages size comparison is broken


sommd:
**Describe the bug**

The size comparasion for similar packages sometimes shows "similar size" for packages that are not at all a similar size.

**To Reproduce**

https://bundlephobia.com/result?p=clipboard@2.0.4

The Gzip size of `clipboard` is 3.2 kB. Under the "Similar Packages" section, `clipboard-copy` (433 B) and `copy-text-to-clipboard` (418 B) are marked as "Similar Size", but `clopboard-polyfill` (2.34 kB) is marked as "26% smaller".

**Expected behavior**

`clipboard-copy` and `copy-text-to-clipboard` should not be marked as "similar size"

This seems to be because the [calculation of `sizeDiff` is completely broken](https://github.com/pastelsky/bundlephobia/blob/103090792d5aedf9f801cfa7ccda61b7115c57ae/client/components/SimilarPackageCard/SimilarPackageCard.js#L33-L34). It's probably better to just pass the `comparisonGzip` to `SimilarPackageCard` rather than trying to recover the size from the percentage.

pastelsky:
Good catch. @sommd Would you be willing to send a PR to fix this?

aviskarkc10:
how about I take this up? might take 1/2 weeks to send in the PR

pastelsky:
@aviskarkc10 you can, sure. The reason behind adding "Similar size" instead of showing the percentage difference for small differences is to avoid giving importance to such trivial differences.

aviskarkc10:
@pastelsky what should the behavior be then once I fix it?

wopian:
Another example is [`axios`](https://bundlephobia.com/result?p=axios@0.19.2) (4.4kb) where it shows `phin` (688b) as a similar size. On `phin`'s page it shows `axios` as 6.6x larger.

PR #288, opened 2020-01-30
Remove condition to display similar size and forbid img tag when purifying dom

aviskarkc10:
1) Remove condition to display similar size #242 
2) Forbid img tag when purifying dom #281


aviskarkc10:
@pastelsky didn't realize I had conflicts. I have fixed 'em

Yedfree (approved):

PR #323, opened 2020-03-18
only block packages in @types org, not anything starting with "@types"

utkarshkukreti:
Right now it blocks this package: https://unpkg.com/@typestrange/decode.


pastelsky:
Thanks for the PR :)

Issue #299, opened 2020-02-18
Include filter in url-encoding for package.json scan results


dawsbot:
**Please describe the feature/suggestion**

The url schema for a `package.json` schema is amazing, I love that I can take this and share it with coworkers. The sorting of Size high -> low does not currently get encoded though, so when a coworker opens up my link, they can't see the same ordering I expected them to

**Describe the solution you'd like**

[...]

PR #326, opened 2020-03-21
includes alphabetical and size filters in url

HarishA97:
Fixes issue #299


pages/scan-results/ScanResults.js
```diff
@@ -192,29 +193,47 @@ class ScanResults extends Component {
...
-  handleSortAlphabetic = () => {
+  setParamsAndState = ( sortMode ) => {
     const { packages } = this.state
-    const sortedList = packages.sort((packA, packB) =>
-      packA.name.localeCompare(packB.name),
-    )
+    const query = packages
+        .map(pack => `${pack.name}@${pack.version}`)
+        .join(',')
+    Router.push(`/scan-results?packages=${query}&sortMode=${sortMode}`)
```
pastelsky:
```suggestion
    Router.replace(`/scan-results?packages=${query}&sortMode=${sortMode}`)
```

We want to replace the history state here so that every sorting change doesn't add a new frame to the history stack.

pages/scan-results/ScanResults.js
```diff
@@ -192,29 +193,47 @@ class ScanResults extends Component {
...
+        const packBSize = packB.result ? packB.result.gzip : 0
+
+        return packBSize - packASize
+      })
+    } else {
+      sortedList = packages.sort((packA, packB) =>
+          packA.name.localeCompare(packB.name),
+      )
+    }
+    return sortedList
```
pastelsky: Nit: A more idiomatic way to do this would be –

```js
  sortBy = (criteria) => (packageA, packageB) => Number 
```

Make `sortBy` return a compare function that can be plugged into line 251 (before).

So the usage in `render` would become –

```js
  packages
    .sort(this.sortBy(this.state.sortMode))
    .map(...)
```

pages/scan-results/ScanResults.js
pastelsky: Also, this approach of building query strings manually can prove to be a little brittle.
Consider what happens as time passes by, and we add more query strings to this page – this function would quickly get out of date.

If the intent of this function is just to update the `sortMode` parameter in URLs, consider constraining it using [PLOP](https://en.wikipedia.org/wiki/Principle_of_least_privilege) (Principle of least privilege).

Then this function would keep all existing query params, and only update what it needs. 

```js
  const updatedQuery =  { ...router.query,  sortMode }
  Router.replace(`/scan-results?${queryString.stringify(updatedQuery)}`)
```

pages/scan-results/ScanResults.js
```diff
@@ -195,29 +197,44 @@ class ScanResults extends Component {
     this.setState({ packages })
   }
 
-  handleSortAlphabetic = () => {
-    const { packages } = this.state
-    const sortedList = packages.sort((packA, packB) =>
-      packA.name.localeCompare(packB.name)
-    )
+  setParamsAndState = ( sortMode ) => {
+    const updatedQuery = { ...Router.query, sortMode}
```
pastelsky, line 201: I think you meant to destructure `this.props.router.query` here.
```suggestion
    const updatedQuery = { ...this.props.router.query, sortMode}
```

PR #333, opened 2020-04-06
In AutocompleteInput, show package name if highlight is not present

### LINKED ISSUE #332: Search autocomplete doesn't always show package names
**Describe the bug**
For some search terms, the search suggestions don't show the package name, but just the package description.

**To Reproduce**
1. Go to https://bundlephobia.com
2. In the package search, type "d3"
3. The package names (d3, d3-scale, ...) will not be shown. It will look like this:
image

**Expected behavior**
The package names are shown.

**Investigation**
This issue happens because npms.io doesn't always return a `highlight` field in its response.
See here: https://api.npms.io/v2/search/suggestions?q=d3 doesn't have that field set.
We should fallback to the package name in that case.

MarcelGerber:
Fixes #332.

This falls back to `package.name` if `highlight` is not present. The item will not get a highlight, but at least it's there 🤷‍♂️

This is akin to how npms.io itself does it: https://github.com/npms-io/npms-www/blob/47e91a387018d27d37e5e1b688369b6cdfdc60da/src/shared/containers/search-box/SearchBox.js#L99-L102

Looks like this (that is, right except without highlighting):
image

### DIFF
--- a/client/components/AutocompleteInput/AutocompleteInput.js
+++ b/client/components/AutocompleteInput/AutocompleteInput.js
@@ -32,7 +32,11 @@ export default class AutocompleteInput extends PureComponent {
         'autocomplete-input__suggestion--highlight': isHighlighted,
       })}
     >
-      <div dangerouslySetInnerHTML={{ __html: item.highlight }} />
+      {item.highlight != null ? (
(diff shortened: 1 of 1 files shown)


pastelsky:
Thanks for the fix @MarcelGerber !

PR #336, opened 2020-04-09
Use JSX's sanitizer for descriptions, drop DOMPurify

### LINKED ISSUE #334: XSS on the site 😅
**Describe the bug**

In my [package.json](https://github.com/fregante/text-field-edit/blob/769732597d8a9c09c2d347c1edf51b48b5921073/package.json#L4):

```json
description: "Insert text in a <textarea> and <input> (supports Firefox and Undo, where possible)"
```

On your site:

bug

**To Reproduce**

Visit: https://bundlephobia.com/result?p=text-field-edit@3.0.0

**Expected behavior**

[Sanitize your outputs!](https://xkcd.com/327/)

MarcelGerber:
Fixes #334.

While XSS is not possible due to the use of DOMPurify, what DOMPurify does is not actually what we want there. If the description contains any HTML, we don't want to display it at all, just show the text. That's why I changed these two up to just use JSX/React's normal HTML escaping.

I've also dropped DOMPurify because we don't need it at the moment.

### DIFF
--- a/client/components/QuickStatsBar/QuickStatsBar.js
+++ b/client/components/QuickStatsBar/QuickStatsBar.js
@@ -1,5 +1,4 @@
 import React, { Component } from 'react'
-import DOMPurify from 'dompurify'
 import './QuickStatsBar.scss'
 
 import TreeShakeIcon from '../../assets/tree-shake.svg'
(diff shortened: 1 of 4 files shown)


pastelsky:
LGTM

PR #358, opened 2020-06-10
Enable dark theme

### LINKED ISSUE #297: Dark theme for websiteDark theme for website

prabuw:
##  High-level description of the approach to enable dark theme:
- Add a react component to capture the theme state.
- Add react-context that to allow the passing of the theme state between the button that toggles the theme and the component that holds the theme state.
- Add `_app.js` ([nextjs docs](https://nextjs.org/docs/advanced-features/custom-app)) to control page initialisation so that the front-end state (e.g. theme choice) persists across pages.
- Toggle a CSS class (`dark-theme`) on the body element that triggers the dark theme.
- Add SASS variables for dark theme colours
- Add styles to components that use the `dark-theme` CSS class to override the normal styles.

## Things considered but not implemented
- Store the user's theming choice to local storage so it is remembered across visits to the site when using the same client and device.

Closes #297.

## Screenshots

image

image

image

image

image


prabuw:
@pastelsky Hello, there was a GitHub issue for implementing a dark theme but I can't seem to find/access it anymore. I wanted to attach this PR to it. It is issue #297 I believe with a label of `good first issue`.

Strangely, GitHub shows there are [3 open issues](https://github.com/pastelsky/bundlephobia/labels) for this label but I can't see them when I click through.

Do you have any idea what has happened? 🤔 😅

VivekNeel:
Awesome, can't wait to get this merged.

piofinn:
Is there any reason why CSS variables aren't used? Seems like it would reduce both the amount and complexity of the code by a lot, and I _highly_ doubt any users of this page use an unsupported browser 😅

Support for global (OS-level) theme and/or remembering the user setting on the page would also be really nice. I'll be happy to assist with any of these if needed! 😊

PR #407, opened 2020-10-01
Improve similar-package-suggestion.md template

dominguezcelada:


pastelsky:
@dominguezcelada  What exactly does your change improve?

dominguezcelada:
> @dominguezcelada  What exactly does your change improve?

Just a better way to express this template's field should be fullfullied with something (assignees in this case) from reporter

Does it makes sense to you?

dominguezcelada:
Close due to inactivity

PR #423, opened 2020-10-31
Update readme with vscode extension

crewsycrews:
Added a VSCode extension with bundlephobia support.


crewsycrews:
@pastelsky any update here? Maybe my extension isn't good enough for now?)

Issue #549, opened 2021-07-21
Incorrect time units for emerging 4G in open graph image


SamChatfield:
**Describe the bug**

There is a bug in the generation of the open graph image for a package result where it will render the incorrect time units for emerging 4G if the time unit for slow 3G is different.

This looks like it's the result of a simple typo on [this line](https://github.com/pastelsky/bundlephobia/blob/28bde79bad73cd1ba0797889f2b77d11aac3424c/utils/draw.utils.js#L201) where it should refer to the `fourGTime.unit` instead of `threeGTime.unit`.

**To Reproduce**

Compare the information rendered by the [frontend for `date-fns-tz@1.1.4`](https://bundlephobia.com/package/date-fns-tz@1.1.4):
<img width="1127" alt="Screenshot 2021-07-21 at 20 18 32" src="https://user-images.githubusercontent.com/4958911/126547086-77135ec7-f828-43ec-8456-64791ade646c.png">

with the information in the [corresponding generated `meta` `og:image`](https://bundlephobia.com/api/stats-image?name=date-fns-tz&version=1.1.4&wide=true):
<img width="1014" alt="Screenshot 2021-07-21 at 20 18 47" src="https://user-images.githubusercontent.com/4958911/126547264-935ebd00-9235-4372-afcd-83a2c9c2c06e.png">

**Expected behavior**

The frontend and open graph images should both display the same time units for emerging 4G.

pastelsky:
Thanks for the report and the PR, will be fixed in our next release.

PR #550, opened 2021-07-21
fix: incorrect time units for emerging 4G in generation of open graph image

SamChatfield:
Small typo fix for the bug I reported in #549.


pastelsky:
Oops! Thanks a lot for sending across this PR.

PR #563, opened 2021-09-15
Use MB for megabytes in 4G speed explanation

danielcompton:
Megabytes is usually written as MB, not mB. I was surprised for a moment when I saw "mB/s" and wasn't quite sure what was being measured.

https://en.wikipedia.org/wiki/Megabyte


pastelsky:
Hi Daniel. Thanks for the PR. 
I've made a few changes to how emerging market speeds are calculated, and 4G no longer remains in the MB/s range (its in KB/s now). 
This PR isn't relevant anymore, and hence closing this.

PR #585, opened 2021-11-22
fix failing ci

amilajack:
this should fix the [broken ci](https://travis-ci.org/github/pastelsky/bundlephobia/jobs/773708564#L238)
Screen Shot 2021-11-22 at 3 42 48 PM


pastelsky:
Travis CI has now been replaced with Github Actions. Thanks for submitting the fix anyway.

PR #596, opened 2021-12-26
Override package install timeout from 30 seconds to 120 seconds

n10l:


Harry-Hopkinson (approved):

n10l:
@pastelsky @Harry-Hopkinson Could you please merge this?

pastelsky:
Why does the package take more than 30 seconds to install though? The reason for the 30s timeout is to protect bundlephobia against resource exhaustion caused by a few packages.

n10l:
@pastelsky Could be due to rollup-plugin-scss which might be causing this while installing node-sass/sass, which I guess compiles on local machine after installation. I am no more working on same thing currently. But when I resume, I will check. But, since few other people were facing this as well https://github.com/pastelsky/bundlephobia/issues/623 I requested for the merge.

PR #679, opened 2022-08-14
feat: support package exports and custom paths

### LINKED ISSUE #218: feature: alternate package entry points
**Please describe the feature/suggestion**

Many packages offer alternate entry points (e.g. `lodash-es/map`, `rxjs/operators`, `@rschedule/core/rules`, etc). At the moment, it appears that bundlephobia only supports scanning they main entry point of a package (e.g. `rxjs`).

It would be awesome if, when searching for packages on `bundlephobia.com`, there was a way of specifying that an alternate entry point should be used. E.g. `rxjs/operators@6.5.3`

### LINKED ISSUE #379: Support for multiple entry points with "exports"
Hello team :wink:

My search fu may be broken but it seems like bundlephobia always expects one single entry point with package.json `"main"` and does not rely on the new `"exports"` feature.

https://nodejs.org/dist/latest-v14.x/docs/api/esm.html#esm_package_entry_points

I have a components library at https://github.com/CleverCloud/clever-components and I would love to see the size report in bundlephobia.

KaelWD:
Requires https://github.com/pastelsky/package-build-stats/pull/52

fixes #218
fixes #379

Next step would be to parse `package.exports` and generate a list of available entrypoints and their sizes, probably have to skip anything with wildcards though.

### DIFF
--- a/client/components/AutocompleteInput/AutocompleteInput.js
+++ b/client/components/AutocompleteInput/AutocompleteInput.js
@@ -72,7 +72,7 @@ export default class AutocompleteInput extends PureComponent {
   render() {
     const { className, containerClass, autoFocus, renderAsH1 } = this.props
     const { suggestions, value, isMenuVisible } = this.state
-    const { name, version } = parsePackageString(value)
+    const { name, version, path } = parsePackageString(value)
     const baseFontSize =
       typeof window !== 'undefined' && window.innerWidth < 640 ? 22 : 35
     const maxFullSizeChars =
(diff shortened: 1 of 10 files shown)


AbhinavAchha (approved):

kurtextrem:
@pastelsky any updates here?

radist2s:
@KaelWD, thank you so much for putting together this PR. 
@pastelsky, this PR is quite important in my opinion, for example, [my library](https://bundlephobia.com/package/@openapi-qraft/react@1.12.0-beta.2) doesn't display the bundle size even though it's done according to the standards. The reason for this is the absence of `module` and `main` in `package.json`, which is not a problem for modern Node.js (>12.x).

PR #690, opened 2022-09-22
refactor: rename to blocklist/allowlist

AndersDJohnson:
Not sure yet if I applied these changes correctly, or if they will work properly, so would likely need help testing & reviewing.

But just thought I'd help start the process of considering a potential rename to more inclusive terms like blocklist and allowlist, as several other projects and teams are doing lately. If it's not feasible to take this on now, or there is disagreement, then these proposed changes can be dismissed.

For reference:
* https://en.wikipedia.org/wiki/Blacklist_(computing)#Controversy_over_use_of_the_term
* https://en.wikipedia.org/wiki/Whitelist#Controversy


pastelsky:
Thanks for your contribution. Could you rebase your PR?

server/middlewares/similar-packages/fixtures.js
```diff
@@ -232,7 +232,7 @@ const categories = {
       { tag: 'escape', weight: Weight.MID },
       { tag: 'filter', weight: Weight.NORMAL },
       { tag: 'xss', weight: Weight.HIGH },
-      { tag: 'whitelist', weight: Weight.SMALL },
+      { tag: 'allowlist', weight: Weight.SMALL },
```
AndersDJohnson, line 235: Perhaps this should remain `tag: 'whitelist'` if it refers to https://www.npmjs.com/search?q=whitelist

server/middlewares/rateLimit.middleware.js
```diff
@@ -4,8 +4,8 @@
 const ipchecker = require('ipchecker')
 const defaults = {
   duration: 1000 * 60 * 60,
-  whiteList: [],
-  blackList: [],
+  allowList: [],
+  blockList: [],
```
AndersDJohnson, line 8: Are these consumer-facing options? Do we need backwards compatibility? @pastelsky

AndersDJohnson:
@pastelsky I have rebased. I left a few comments with some questions before we should proceed though. How can we test this well?

PR #703, opened 2022-10-24
add checksumBehavior with update value to allow installing dependencies with yarn

raulmarindev:
By default, Yarn will [throw an exception](https://yarnpkg.com/configuration/yarnrc) on yarn install if it detects that a package doesn't match the checksum stored within the lockfile. This was the change applied:

1.  Open your .yarnrc.yml.
2.  Add the following configuration option: ` checksumBehavior: "update"`


pastelsky:
Closing this as the checksum has now been updated.

PR #704, opened 2022-10-30
migrate to next 13

darasus:
**What?**
- upgrade to Nextjs 13
- move all sass imports to `stylesheets` folder and moving imports to `_app`
- remove unnecessary `postcss` config
- remove unnecessary next `plugins`

**Why?**
Breaking changes while migrating from Nextjs 11 to 13.

**Tested?**
- local


pastelsky:
> move all sass imports to stylesheets folder and moving imports to _app
Why has this been done? I'd rather have the styles for components

darasus:
@pastelsky fair point. What I did in the last commit I reverted component scss file locations except for page level sass, this can't stay in pages directory.

pages/compare/ComparePage.js
```diff
@@ -118,7 +117,7 @@ export default class ResultPage extends PureComponent {
         <div className="page-container">
           <header className="result-header">
             <section className="result-header--left-section">
-              <Link href="/">
+              <Link href="/" legacyBehavior>
```
pastelsky, line 120: Should we consider removing the `a` tag completely?

package.json
```diff
@@ -115,19 +112,16 @@
...
     "@babel/core": "^7.17.10",
     "@babel/preset-env": "^7.17.10",
-    "@zeit/next-source-maps": "^0.0.4-canary.1",
     "autoprefixer": "^9.8.6",
     "babel-jest": "^24.9.0",
     "babel-plugin-css-modules-transform": "^1.6.1",
     "babel-plugin-inline-react-svg": "^1.1.1",
     "babel-plugin-module-resolver": "^4.0.0",
     "babel-plugin-wrap-in-js": "^1.1.1",
+    "eslint-config-next": "^13.0.0",
```
pastelsky, line 124: Why has this been added?

pastelsky:
Why have all components been moved from their organised page directories to a flat folder? This makes things quite unmaintainable. What was next complaining about here?

next.config.js
```diff
@@ -1,19 +1,9 @@
-const withSass = require('@zeit/next-sass')
-const withCss = require('@zeit/next-css')
-const withSourceMaps = require('@zeit/next-source-maps')()
+const path = require('path')
 
 module.exports = {
-  webpack5: false,
-  pageExtensions: ['page.js'],
```
pastelsky, line 7: Avoid removing this — doing this is going to break pages that use the `index.page.js` suffix. e.g. The `/scan` route.

pastelsky (changes requested):

package.json
darasus: I believe it's was a postinstall script that installed this. I can try remove it see if build fails.

darasus:
> Why have all components been moved from their organised page directories to a flat folder? This makes things quite unmaintainable. What was next complaining about here?

- first of all catch all route is now only allowed as last segment
- besides I thought it would make it much more clear of what is a page inside `pages` directory if we just leave only pages there. What is the reason to keep components inside pages while we have root level `components` folder? Technical all root level components are used by pages therefore I don't see the need to keep another set of components inside pages directory. The flatter the folder/file tree the better no?

pages/compare/ComparePage.js
darasus: We can yes, will fix

pastelsky:
> don't see the need to keep another set of components inside pages directory

The top level `components` directory contains re-useable components, while the ones inside the page's directory are specifically for use in the page. 

Let's please make the minimum amount of changes required to upgrade to Next 13. Changing the entire code hierarchy in a single PR is un-necessary, hard to review and quite confusing — these need to be discussed before a change is considered.

darasus:
>

darasus:
> > don't see the need to keep another set of components inside pages directory
> 
> The top level `components` directory contains re-useable components, while the ones inside the page's directory are specifically for use in the page.
> 
> Let's please make the minimum amount of changes required to upgrade to Next 13. Changing the entire code hierarchy in a single PR is un-necessary, hard to review and quite confusing — these need to be discussed before a change is considered.

Agreed, should have been discussed first.

darasus:
@pastelsky I believe I considered all your comments above, please have a look. Please pay attention that we use webpack5 now and your override is not present anymore. What kind of patch did you have there?

pastelsky:
I still think there are quite a few un-related changes made in this PR — kindly only keep the minimum change required to upgrade to Next 13, and undo everything else — raise a separate issue / PR if needed for those.

darasus:
@pastelsky could you give an example of what change is not relevant to next 13 migration?

pastelsky:
@darasus I've added comments for the same  — could you check the "pending" comments from the latest review?

next.config.js
darasus: I have reverted `pageExtensions: ['page.js']` but webpack 4 is not supported anymore.

pages/compare/ComparePage.js
darasus: This is fixed

package.json
darasus: Removed this package

darasus:
@pastelsky I believe all pending comments should be resolved now, basically removed eslint package that was installed by next's postinstall

pastelsky:
@darasus There are open comments that haven't been responded too :) 
I'm listing them down again here to make it easier —
1. We don't need to move page-specific components to the top-level components directory. I'm happy to rename the top level components directory to `shared-components` in case its confusing. Please undo these (e.g. InterlinksSection, ExportsAnalysis)
2. A bunch of CSS has been deleted from `client/components/Stat/Stat.scss` — I'm not entirely sure why
3. `resolutions` key from package.json has been deleted — this shouldn't be deleted
4. `pages/package/[...packageString]/index.page.js` (and a few other page components) have been renamed to `pages/package/[...packageString].page.js` — this is un-necessary. By convention we nest all pages under a directory so we don't need to refactor them in case there are more components to be added in the same directory.

darasus:
@pastelsky thanks for summarising you comments!

1. This is not the case anymore, I've reverted these in the previous commits.
2. Not sure what happened with `Stat.scss`, will revert it
3. Reverted resolutions
4. And reverted pages

How is it looking now?

pastelsky:
Thanks for fixing all of the issues. This looks good to me. 

As a next step, we can —
1. Cleanup bundlephobia's `babelrc` so that Next JS can start using `swc` for compilation
2. Think about using React server components to make bundlephobia even faster and fix SSR

pastelsky (approved):

pastelsky:
Congratulations on your first successful contribution to bundlephobia.

PR #706, opened 2022-11-01
chore: enable eslint

darasus:
Enable ESLint and fix errors

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,6 @@
+{
+  "extends": "next",
+  "plugins": [],
+  "root": true,
+  "rules": {}
+}
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,5 +1,5 @@
 {
-  "extends": "next",
+  "extends": ["next", "prettier"],
   "plugins": [],
   "root": true,
   "rules": {}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,6 +24,8 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
       - run: yarn install
+      - name: Run lint
+        run: yarn next lint
       - run: yarn build
         env:
           CI: true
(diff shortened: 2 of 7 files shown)


pastelsky:
Thanks for submitting this change. We should also add this to —
1. Autofix ESLint errors using the current husky precommit hook
2. Add ESLint checks to our CI (`.github/workflows/ci.yml`)
3. Add a lint script to `package.json`
4. Enable ESLint prettier (https://prettier.io/docs/en/integrating-with-linters.html) so formatting changes are linted as well.

darasus:
@pastelsky I think all comments are taken care of. Mind having another look?

.github/workflows/ci.yml
```diff
@@ -24,6 +24,8 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
       - run: yarn install
+      - name: Run lint
+        run: yarn next lint
```
pastelsky, line 28: Why do we have to call "next lint"? What is next providing here additional to what ESLint already provides? 
If its not much, I'd recommend not adding a dependency on `next` for ESLint.

pastelsky (commented):
LGTM, except a question about why we're coupling lint with `next`.

.eslintrc.json
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": ["next", "prettier"],
```
pastelsky, line 2: We should add `eslint:recommended` as well here (plus a few more plugins). "next", "prettier" are specific rulesets for frameworks.

Happy for this to be done in a follow up PR though

darasus: I'll do that here in one go

darasus: Or maybe not, `eslint:recommended` is much more strict, will do that in the following PR

pastelsky (approved):

darasus:
@pastelsky thanks for approving but i don't have permissions to merge.

PR #708, opened 2022-11-02
Kickoff TypeScript migration

igordanchenko:
1) Add Typescript dependencies
2) Migrate blog pages and components to .tsx


.gitignore
```diff
@@ -14,5 +14,10 @@ node_modules
 yarn-error.log
 logs
 **/.yarn/cache
+**/.yarn/install-state.gz
 
+# JetBrains IDE
+.idea
 
+# TypeScript
+next-env.d.ts
```
darasus, line 23: I don't think this should be gitignored

darasus:
Based on my experience with PR reviews in this repository I'd suggest to keep scope of this PR to bare minimum. If you're integrating typescript then I would reset actual migration changes (maybe only keep it to a single file to validate that it's working fine).

pages/blog/components/Article.tsx
```diff
@@ -16,24 +16,28 @@ const ArticleWithContent = () => {
...
     return (
       <pre>
+        {/* @ts-ignore */}
         <code>{error}</code>
       </pre>
     )
   } else if (data) {
     return (
       <>
+        {/* @ts-ignore */}
```
darasus, line 40: I think `ts-ignore` should be avoided at all costs, I'd rather have `any`.

pages/blog/components/Article.tsx
```diff
@@ -16,24 +16,28 @@ const ArticleWithContent = () => {
 }
 
 const Article = () => {
-  const { router } = useRouter()
+  const router = useRouter()
 
   console.log('router is ', router)
+
+  // ts-ignoring react-contentful type errors for now because their type definitions are too generic
```
pastelsky, line 23: Shouldn't this be a `@ts-ignore`?

pages/blog/components/Article.tsx
pastelsky: What is the TS Error here? If the library doesn't have a good definition available, we can provide "good enough" definitions through casting.

pages/blog/index.page.tsx
```diff
@@ -27,23 +28,23 @@ const BlogHome = () => {
...
     content = (
       <pre>
+        {/* @ts-ignore */}
         <code>{error}</code>
       </pre>
     )
   } else if (data) {
     content = (
       <>
+        {/* @ts-ignore */}
```
pastelsky, line 38: Let's type and re-export `data.items` instead of littering our codebase with ignores.

.gitignore
pastelsky: This does need to be ignored as per —
https://nextjs.org/docs/basic-features/typescript

pastelsky (commented):
LGTM, pending typing the contentful `data` object (explicitly) so we don't end up adding a bunch of TS ignores.

pages/blog/components/Article.tsx
igordanchenko: > What is the TS Error here?

"TS2339: Property 'items' does not exist on type 'object'". Basically both "data" and "error" are typed as object in react-contentful types... 😞 

```tsx
export interface HookResponse {
  data?: object;
  error?: object;
  fetched: boolean;
  loading: boolean;
}
```

> If the library doesn't have a good definition available, we can provide "good enough" definitions through casting.

I wasn't sure what would be your preferred approach and thought it would be better to discuss this first (hence, the `@ts-ignore` "for now"). In this case the library exposes the return type as an interface, so we can tweak it to our needs via module augmentation. Does this sound good?

igordanchenko: @pastelsky, please take another look once you get a chance. I went ahead and augmented `react-contentful` types, as well as removed all `@ts-ignore`s

pages/blog/index.page.tsx
igordanchenko: Done.

pastelsky (approved):
LGTM

igordanchenko:
Type checking runs as part of `yarn build` - which is also executed in GitHub CI workflow. 

Are you sure you want a separate script for this?

pastelsky:
Gotcha! Good to go then.

pastelsky (approved):

PR #710, opened 2022-11-02
TypeScript migration - client components

igordanchenko:
Migrate client/components to TypeScript


client/components/Layout/Layout.tsx
```diff
@@ -1,13 +1,20 @@
...
+
+type LayoutState = {
+  recentSearches: string[]
+}
 
-export default class Layout extends Component {
+export default class Layout extends Component<LayoutProps, LayoutState> {
   state = {
-    recentSearches: [],
+    recentSearches: [] as string[],
```
darasus, line 17: Is casting here necessary?

client/analytics.ts
```diff
@@ -46,41 +64,71 @@ export default class Analytics {
...
       timeTaken,
     })
   }
 
-  static exportsSizesFailure({ packageName, timeTaken }) {
+  static exportsSizesFailure({
+    packageName,
+    timeTaken,
+  }: {
+    packageName: string
```
darasus, line 129: You could avoid duplication by having a type like:

```
type HasPackageName = { packageName: string }
type HasTimeTaken = { timeTaken: number }
```

And then intersect those together and compose argument types on each method

types/index.ts
```diff
@@ -0,0 +1,12 @@
...
+  description: string
+  repository: string
+  dependencyCount: number
+  isTreeShakeable: boolean
+  hasSideEffects: string[] | boolean
+}
+
+export type PropsWithClassName<P = unknown> = P & {
+  className?: string | undefined
+}
```
darasus, line 12: I'm not sure I like this type, it's a bit too high level. Instead I would do something like this:

```
type Maybe<T> = T | null | undefined
type HasClassName = {
  className: string;
}
types Props = React.PropsWithChildren<Maybe<HasClassName>>
```

client/analytics.ts
igordanchenko: Good point. I'll revise.

types/index.ts
igordanchenko: Doesn't it serve the same purpose as React.PropsWithChildren?

darasus: Yes I meant this `types [Component]Props = React.PropsWithChildren<Maybe<HasClassName>>` goes into your component.

igordanchenko: What I meant here is that `PropsWithClassName` serves very similar purpose to what `React.PropsWithChildren` does, and it's implemented in the same way. Additionally it allows to use it as a wrapper as well as an intersection:

```tsx 
type Props1 = React.PropsWithChildren<
  PropsWithClassName<{
    footer?: React.ReactNode
  }>
>

type Prop2 = React.PropsWithChildren &
  PropsWithClassName & {
    footer?: React.ReactNode
  }
```

vs the type you suggested can be used only as an intersection:

```tsx
type Props3 = React.PropsWithChildren<Maybe<HasClassName>> & {
  footer?: React.ReactNode
}
```

darasus: But in my case you have greater control, let's say in one edge case you want to enforce presence of `className` prop? Also I don't see anything wrong with intersections, why is that bad?

darasus: Nothing prevents you from having a global `PropsWithClassName` defined as:

```
type PropsWithClassName<P = unknown> = P & Maybe<HasClassName>
```

By composing types with smaller units you have greater flexibility. The only problem that I'm seeing is that `PropsWithClassName` is very limiting by its definition and couples entire component folder to a type that might change thus making harder introducing changes to a specific component. But again it's just my recommendation.

igordanchenko: Well, I must have paid closer attention to your proposal in the first place as it's functionally incorrect. `Maybe<HasClassName>` resolves to `{ className: string } | null | undefined` which is not the same thing as `{ className?: string }`

igordanchenko: > The only problem that I'm seeing is that PropsWithClassName is very limiting by its definition and couples entire component folder to a type that might change thus making harder introducing changes to a specific component.

IMO, nothing wrong with having project-wide utilities. The person making the change will need to make sure they are not breaking stuff.

darasus: > Well, I must have paid closer attention to your proposal in the first place as it's functionally incorrect. Maybe<HasClassName> resolves to { className: string } | null | undefined which is not the same thing as { className?: string }

Not the same signature yes but end type safety is the same.

> IMO, nothing wrong with having project-wide utilities. The person making the change will need to make sure they are not breaking stuff.

Very true.

igordanchenko: > Not the same signature yes but end type safety is the same.

Absolutely not. The following code doesn't even compile.

```tsx
type Props = Maybe<HasClassName>

const Component = ({ className }: Props) => {}
```

> TS2339: Property 'className' does not exist on type 'Props'.

because the `Props` type is effectively `{ className: string } | null | undefined`

darasus: Am I missing something? https://stackblitz.com/edit/react-ts-tnvc8b?file=App.tsx

igordanchenko: I wouldn't 100% trust online tools like Stackblitz or CodeScandbox when it comes to type checking.

Just try pasting that code into bundlephobia project and see what `tsc` tells you.

```shell
 % tsc
client/components/HasClassNameTest.tsx:8:22 - error TS2339: Property 'className' does not exist on type 'Props'.

8 const Component = ({ className }: Props) => {
                       ~~~~~~~~~


Found 1 error in client/components/HasClassNameTest.tsx:8
```

Also try removing `className` prop here:

```tsx
<Component />
```

darasus: It's because we have `"strict": true`. If you want to keep that setting you can avoid this by not destructuring:

```
const Component = (props: Props) => {
return <div className={props?.className} />
}
```

Or you can do `"strictNullChecks": false` then you can destructure.

Another option is to use `Partial`, then it accepts existence of the containing object just makes properties optional.

```
type Props = Partial<HasClassName>
```

igordanchenko: > It's because we have "strict": true

And that is the default that `tsc --init` creates.

> Or you can do "strictNullChecks": false then you can destructure.

Explain it to me like I'm 5... why would I want that considering

> When strictNullChecks is false, null and undefined are effectively ignored by the language. **This can lead to unexpected errors at runtime.**

igordanchenko: > Another option is to use Partial, then it accepts existence of the containing object just makes properties optional.

So you are proposing we type `type Props = Partial<HasClassName>` 99% of the time just to be able to type `type Props = HasClassName` in that 1 case when the `className` should actually be required? I'm not buying it as I honestly do not see a point. I'd much rather prefer to type `type Props = PropsWithClassName` 99% of the time and `type Props = Required<PropsWithClassName>` in that 1 potential case.

darasus: Sorry I didn't mean it to turn it such a long discussion, really feel free to ignore my comments on this.

But :) I'd try to be more specific. What I really don't like is that your type `PropsWithClassName` takes in generic. I think better approach is intersection.

```
// /types.ts
type HasClassName = { className: string }
type HasClassNameMaybe = Partial<HasClassName>
```

And then use that in your component as:

```
type Props = React.PropsWithChildren<HasClassNameMaybe & SomeOtherType>
```

versus:

```
type Props = React.PropsWithChildren<PropsWithClassName<SomeOtherType>>
```

You approach is like functional programming, more confusing :D And requires to see the implementation of you type to understand what happens with my argument. Why intersection is more straightforward and simply implies concatenation.

Besides what about other cases of reused types. Are you planning to name them all `PropsWithSomething<T>`?

igordanchenko: > Your approach is like functional programming, more confusing :D

I see, so it's just a matter of personal preference with a sprinkle of not-invented-here 😉 

> And requires to see the implementation of your type to understand what happens with my argument. While intersection is more straightforward and simply implies concatenation.

Well I honestly thought this type will be trivial to digest for anyone familiar with `React.PropsWithChildren`, but evidently I was wrong. 

Do you also disapprove using `React.PropsWithChildren` as a wrapper

```tsx
type Props = React.PropsWithChildren<{
  some: prop;
}>
```

but require to intersect it instead? (I'm genuinely curious)

```tsx
type Props = React.PropsWithChildren & {
  some: prop;
}
```

Will you be more comfortable with the following implementation?

```tsx
export type WithClassName = {
  className?: string
}

type Props = React.PropsWithChildren & WithClassName & SomeOtherType;
```

Let me elaborate on the type name right off the bat... I won't name it `PropsWithClassName` because it would be too similar to `React.PropsWithChildren` while they can't be used in a similar way. I'm also not naming it `HasClassName` because I'm assuming you prefer to use "Has" prefix for required props. I'm also not using "Maybe" prefix / suffix / wrapper because it's redundant as the prop is optional in 100% of the cases.

darasus: I'm not approving or disapproving anything here, I'm just recommending 🙂

```
export type WithClassName = Pick<HTMLAttributes<HTMLElement>, 'className'>;

type Props = React.PropsWithChildren & WithClassName & SomeOtherType;
```
This is slightly over engineered type 🙂 but I think I prefer it much more except for how you use it. I think this is my favorite:

```
export type WithClassName = Pick<HTMLAttributes<HTMLElement>, 'className'>;

type Props = React.PropsWithChildren<WithClassName & SomeOtherType>;
```
Whether it's `WithClassName` or `HasClassName` I don't care, both are fine by me.

I'm only against doing this:

```
type Props = React.PropsWithChildren<PropsWithClassName<PropsWithSomethingElse<SomeOtherType>>>;
```

igordanchenko:
@pastelsky do you have any feedback on this PR?

pastelsky (changes requested):

types/index.ts
```diff
@@ -0,0 +1,12 @@
+export type PackageInfo = {
+  name: string
+  description: string
+  repository: string
+  dependencyCount: number
+  isTreeShakeable: boolean
+  hasSideEffects: string[] | boolean
+}
+
+export type PropsWithClassName<P = unknown> = P & {
```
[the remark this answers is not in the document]
igordanchenko: You are looking at the outdated code that is no longer part of this PR. Please review the up-to-date version of the code.

client/components/AutocompleteInputBox/AutocompleteInputBox.tsx
```diff
@@ -1,7 +1,15 @@
...
 import cx from 'classnames'
 
-class AutocompleteInputBox extends Component {
+import { PropsWithClassName } from '../../../types'
+
+type AutocompleteInputBoxProps = React.PropsWithChildren<
+  PropsWithClassName<{
+    footer?: React.ReactNode
+  }>
+>
```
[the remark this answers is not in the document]
igordanchenko: You are looking at the outdated code that is no longer part of this PR. Please review the up-to-date version of the code.

client/components/QuickStatsBar/QuickStatsBar.tsx
```diff
@@ -74,7 +84,9 @@ class QuickStatsBar extends Component {
           <div className="quick-stats-bar__stat">
             <SideEffectIcon className="quick-stats-bar__stat-icon" />{' '}
             <span>
-              {hasSideEffects.length ? 'some side-effects' : 'side-effect free'}
+              {!(hasSideEffects === false)
```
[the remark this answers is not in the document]
igordanchenko: Fixed the logic. Please review.

types/amplitude.d.ts
```diff
@@ -0,0 +1,10 @@
+// stab typings for amplitudeScript loaded in the _document
```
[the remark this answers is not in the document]
igordanchenko: Fixed

client/components/ProgressHex/progress-hex-timeline.ts
```diff
@@ -17,22 +33,22 @@ export default class ProgressHexAnimator {
     })
 
     Array.from(this.circles).forEach(circle => {
-      const cx = parseFloat(circle.getAttribute('cx'))
-      const cy = parseFloat(circle.getAttribute('cy'))
+      const cx = parseFloat(circle.getAttribute('cx')!)
```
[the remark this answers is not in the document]
igordanchenko: Couple options:

1) Always provide fallback values, but these fallback values will **never** be used as the attributes in question are always set in the upstream code.

2) Replace direct `getAttribute` calls with a type-safe wrapper (but again, the `if` clause will never be executed):

```
const getRequiredAttribute = (
  element: Element,
  attribute: string,
  fallback?: string
) => {
  const value = element.getAttribute(attribute) ?? fallback
  if (value === null) {
    throw new Error(`Attribute '${attribute}' is missing`)
  }
  return value
}
```

All in all, I'm not sure these "!" windmills are worth fighting (*in this particular case*). This is not a public API and there is no goal to protect the code against incorrect usage. The `map.get` calls that are closer to the end of that file would be even more tedious to fix.

What would be your preferred approach here?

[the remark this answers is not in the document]
pastelsky: Makes sense in this case — we don't expect this to be `undefined`, and even if it is, the damage is quite limited. I think we can make a case by case decision here — I'm generally a big fan of `invariant` / `nullthrows`

client/components/AutocompleteInputBox/AutocompleteInputBox.tsx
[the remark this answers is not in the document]
pastelsky: Ah, so this comment was added much earlier but I didn't click on request changes, which is why it never made it to the PR until an hour back. Ignore this. This part of github's UX does throw me off quite a bit since I'm used to bitbucket.

client/components/QuickStatsBar/QuickStatsBar.tsx
```diff
@@ -74,7 +84,9 @@ class QuickStatsBar extends Component {
           <div className="quick-stats-bar__stat">
             <SideEffectIcon className="quick-stats-bar__stat-icon" />{' '}
             <span>
-              {hasSideEffects.length ? 'some side-effects' : 'side-effect free'}
+              {!(hasSideEffects === false) && hasSideEffects.length
```
pastelsky, line 87: Nit — this can simply be `hasSideEffects && hasSideEffects.length` perhaps?

pastelsky (approved):

pastelsky:
@igordanchenko This is lovely — thanks again for taking the time out to help us move to typescript. I wonder how much remains after this — I guess not a lot.

client/components/QuickStatsBar/QuickStatsBar.tsx
igordanchenko: You are correct.

igordanchenko:
@pastelsky Client components are pretty much done, but I want to say there is still a fair amount of work left - most of the server-side code and pages.

PR #713, opened 2022-11-03
ts-ify and refactor AutocompleteInput

darasus:
- ts-ify `AutocompleteInput`
- break down into smaller components and hooks


client/components/AutocompleteInput/AutocompleteInput.tsx
```diff
@@ -0,0 +1,121 @@
...
+}
+
+export const AutocompleteInput = ({
+  initialValue = '',
+  renderAsH1 = false,
+  className,
+  containerClass,
+  autoFocus,
+  onSearchSubmit,
+}: Props) => {
```
raulmarindev, line 28: I use this exact same approach: interface for props and Props for name to avoid redundancy ❤. Just in case, though, I wanted to bring to your attention that @igordanchenko is using a different pattern: type alias and [Component]Props name. For consistency purposes it would be amazing if an agreement could be met regarding which of the two should be used going forwards. @pastelsky

igordanchenko: 2 reasons for type + [Component]Props pattern:

1) I never ran into a need to use declaration merging when it comes to component props, hence the "type" was always sufficient

2) [Component]Props pattern allows to have more than one component per file would the need be

If I'm not mistaken I picked up this pattern from one of the Kent's articles - https://kentcdodds.com/blog/how-to-write-a-react-component-in-typescript

darasus: While I'm happy to follow the style that majority of you follow I still don't see enough of a reason to go the way you suggest. How often you have multiple components in one file? I'd argue this is worse since you loose these components in file search. You only add more boilerplate with repeating component names. Wether it's type or interface I wouldn't care that much, types are better as they allow unions. Anywho I'll adjust according to your feedback no prob.

igordanchenko: > How often you have multiple components in one file?

Pretty rarely, but one time is enough to break the "interface Props" pattern.

> I'd argue this is worse since you loose these components in file search.

I entirely agree, but there are cases when it may be reasonable. For example, split up a large plain component into several logical "sub-components" that will never be re-used and do not necessarily "deserve" their own file.

> You only add more boilerplate with repeating component names.

Yes, it may seem like boilerplate, but IMO it's more descriptive than  `import {Props as MyComponentProps} from "./MyComponent"` if you ever need to reference props type elsewhere.

Anyway, I'm just explaining where I come from with respect to the pattern I'm using. I'm not the one who brought this subject up, and I'm not pushing for a unified style guide at this point, as the codebase could see much love in so many other aspects before we start arguing about nitty-gritty aspects of code style (again, IMO).

raulmarindev: > Anyway, I'm just explaining where I come from with respect to the pattern I'm using. I'm not the one who brought this subject up, and I'm not pushing for a unified style guide at this point, as the codebase could see much love in so many other aspects before we start arguing about nitty-gritty aspects of code style (again, IMO).

I may understand where you are coming from but there's no time like the beginning to decide on common styles. This "nitty-gritty" stuff as you call it is very important when working on a team. From my understanding, the app has remained unchanged for a long time. and I believe there is no need to rush forward. Then again, I'd appreciate @pastelsky opinion on this.

darasus: > Pretty rarely, but one time is enough to break the "interface Props" pattern.
But isn't it better to use preferred pattern and break it couple times than the one designed for a minor edge case?

> I entirely agree, but there are cases when it may be reasonable. For example, split up a large plain component into several logical "sub-components" that will never be re-used and do not necessarily "deserve" their own file.

My rule is that moving components to local components folder with the idea that it's going to be used only once within the scope of the parent directory is a good pattern. You reduce a large chunk of cognitive load by having smaller files, you don't need to think/read where is this export keyword, what is actually being consumed here by outside...

> Yes, it may seem like boilerplate, but IMO it's more descriptive than import {Props as MyComponentProps} from "./MyComponent" if you ever need to reference props type elsewhere.

This `import {Props as MyComponentProps} from "./MyComponent"` must never happen, if you want to share type with the outside, move it to closest shared `types.ts`. Or export it from the entry point `index.ts` and rename it there (this approach is least preferred by me but sometimes makes sense in the feature folder pattern when you need to have independent features).

```
// components/Avatar.tsx

type Props = SharedProps;
```

Honestly I'm arguing here just for fun :) But I completely agree with your last statement. Let's move on :)

darasus: Updated in the last commit

igordanchenko:
> But isn't it better to use preferred pattern and break it couple times than the one designed for a minor edge case?

My OCD isn't sitting well with this 😆 

> Honestly I'm arguing here just for fun :) But I completely agree with your last statement. Let's move on :)

🍻

client/components/AutocompleteInput/AutocompleteInput.tsx
igordanchenko: > But isn't it better to use preferred pattern and break it couple times than the one designed for a minor edge case?

My OCD isn't sitting well with this 😆

> Honestly I'm arguing here just for fun :) But I completely agree with your last statement. Let's move on :)

🍻

client/components/AutocompleteInput/hooks/useAutocompleteInput.ts
```diff
@@ -0,0 +1,57 @@
+import React from 'react'
+import debounce from 'debounce'
+
+import { parsePackageString } from '../../../../utils/common.utils'
+import API from '../../../api'
```
raulmarindev, line 5: Perhaps it could be a good idea to enable absolute imports at some point 😁

darasus: I'd argue if absolute imports are a good idea, you can't enforce them in different IDEs. Unless there is a way to do that? I prefer monorepo approach with named package aliases with a single possible entry point.

pages/package/[...packageString]/ResultPage.js
```diff
@@ -3,7 +3,7 @@ import Analytics from '../../../client/analytics'
 
 import ResultLayout from '../../../client/components/ResultLayout'
 import BarGraph from '../../../client/components/BarGraph'
-import AutocompleteInput from '../../../client/components/AutocompleteInput'
+import { AutocompleteInput } from '../../../client/components/AutocompleteInput/AutocompleteInput'
```
Comment (pastelsky) [inline pages/package/[...packageString]/ResultPage.js:6]: Let's use a `index.ts` re-export to keep the imports clean. 
`components/AutocompleteInput/AutocompleteInput` reads odd

client/components/AutocompleteInput/hooks/useFontSize.ts
```diff
@@ -0,0 +1,18 @@
...
+    const searchFontSize =
+      value.length < maxFullSizeChars
+        ? null
+        : `${baseFontSize - (value.length - maxFullSizeChars) * 0.8}px`
+
+    return searchFontSize
+  }, [value])
+
+  return { searchFontSize }
+}
```
pastelsky, line 18: Nice!

client/components/AutocompleteInput/components/PackageNameElement.tsx
```diff
@@ -0,0 +1,7 @@
+interface Props extends React.HTMLAttributes<HTMLElement> {
+  as?: 'h1'
+}
+
+export function PackageNameElement({ as, ...props }: Props) {
+  return as === 'h1' ? <h1 {...props} /> : <span {...props} />
+}
```
pastelsky, line 7: nit / optional:  I'm not a big fan of splitting components into new files for the sake of it. If a component is small enough, and will only ever be used in one component, keeping it colocated avoids unnecessary context switching.

client/components/AutocompleteInput/components/PackageNameElement.tsx
```diff
@@ -0,0 +1,7 @@
+interface Props extends React.HTMLAttributes<HTMLElement> {
+  as?: 'h1'
```
pastelsky, line 2: nit: If this is to only have one value, it is better modelled as a boolean.
`isHeading` (for e.g.)

pastelsky (changes requested):
Looks good pending a few minor/optional comments. Thanks for the submission!

pages/package/[...packageString]/ResultPage.js
Comment (darasus) [reply pages/package/[...packageString]/ResultPage.js]: ✅

client/components/AutocompleteInput/components/PackageNameElement.tsx
darasus: ✅

client/components/AutocompleteInput/components/PackageNameElement.tsx
darasus: ✅

darasus:
@pastelsky could you take a look again?

pastelsky:
LGTM pending conflict resolution

darasus:
@pastelsky all resolved, take a look

pastelsky (approved):

pastelsky:
LGTM, pending local testing

Issue #684, opened 2022-08-21
If the package version is long, it overflows the interface


cawa-93:
**Describe the bug**
![зображення](https://user-images.githubusercontent.com/1662812/185790949-eb2e80c4-67cf-4334-b3ec-3de61d64c555.png)

**To Reproduce**
Go to https://bundlephobia.com/package/@iconify-prerendered/vue-mdi

PR #714, opened 2022-11-03
fix long bar version

darasus:
- fix long bar version ([issue](https://github.com/pastelsky/bundlephobia/issues/684))

CleanShot 2022-11-03 at 08 21 35@2x

### DIFF
--- a/client/components/BarVersion/BarVersion.scss
+++ b/client/components/BarVersion/BarVersion.scss
@@ -1,0 +1,46 @@
+@import '../../../stylesheets/colors';
+@import '../../../stylesheets/variables';
+
+.bar-graph__bar-version {
+  @include font-size-xs;
+  z-index: 33;
+  font-weight: $font-weight-light;
+  transform: rotate(-180deg);
+  font-variant-numeric: tabular-nums;
+  color: lighten($raven, 15%);
+  transition: opacity 0.2s, color 0.2s;
+  font-family: $font-family-code;
+  letter-spacing: -1px;
+  line-height: 1;
+  cursor: pointer;
+  height: $bar-width;
+  writing-mode: vertical-rl;
+  text-orientation: sideways;
+  height: 50px;
+  text-align: left;
+  justify-content: flex-start;
+  display: inline-block;
+  margin-top: 5px;
+
+  > div {
+    display: flex;
+    text-align: end;
+    justify-content: flex-start;
+    align-items: center;
+    height: 100%;
+    width: 100%;
+    > span {
+      white-space: nowrap;
+      overflow: hidden;
+      text-overflow: ellipsis;
+    }
+  }
+
+  .bar-graph-container:hover & {
+    color: $raven;
+  }
+
+  .bar-graph__bar-group:hover & {
+    color: darken($raven, 20%);
+  }
+}
--- a/client/components/BarVersion/BarVersion.scss
+++ b/client/components/BarVersion/BarVersion.scss
@@ -5,7 +5,7 @@
   @include font-size-xs;
   z-index: 33;
   font-weight: $font-weight-light;
-  transform: rotate(-180deg);
+  transform: rotate(-90deg) translateX(#{-$global-spacing * 1.5});
   font-variant-numeric: tabular-nums;
   color: lighten($raven, 15%);
   transition: opacity 0.2s, color 0.2s;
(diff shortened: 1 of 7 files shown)


client/components/BarVersion/BarVersion.scss
```diff
@@ -0,0 +1,46 @@
...
+    display: flex;
+    text-align: end;
+    justify-content: flex-start;
+    align-items: center;
+    height: 100%;
+    width: 100%;
+    > span {
+      white-space: nowrap;
+      overflow: hidden;
+      text-overflow: ellipsis;
```
pastelsky, line 35: The ellipsis part doesn't seem to work in Safari.

client/components/BarVersion/BarVersion.scss
```diff
@@ -0,0 +1,46 @@
...
+  > div {
+    display: flex;
+    text-align: end;
+    justify-content: flex-start;
+    align-items: center;
+    height: 100%;
+    width: 100%;
+    > span {
+      white-space: nowrap;
+      overflow: hidden;
```
pastelsky, line 34: Let's add a title attribute so folks can hover over to see the full version if they wish to.

client/components/BarVersion/BarVersion.scss
darasus: CleanShot 2022-11-03 at 16 01 52@2x

Looks okay to me?

darasus: Also here seems to have support https://caniuse.com/?search=ellipsis

client/components/BarVersion/BarVersion.scss
```diff
@@ -0,0 +1,46 @@
...
+  color: lighten($raven, 15%);
+  transition: opacity 0.2s, color 0.2s;
+  font-family: $font-family-code;
+  letter-spacing: -1px;
+  line-height: 1;
+  cursor: pointer;
+  height: $bar-width;
+  writing-mode: vertical-rl;
+  text-orientation: sideways;
+  height: 50px;
```
pastelsky, line 19: There's already a height attribute above (on that note, we should probably add stylelint as well to this project).

client/components/BarVersion/BarVersion.scss
```diff
@@ -0,0 +1,46 @@
+@import '../../../stylesheets/colors';
+@import '../../../stylesheets/variables';
+
+.bar-graph__bar-version {
```
pastelsky, line 4: This solution produces a weird alignment for a package like `react-pdf@4.2.0-beta`

Screenshot 2022-11-03 at 8 32 04 PM

Ideally all of these versions would stick up to the top? Arguable.

client/components/BarVersion/BarVersion.scss
pastelsky: You need to test with a package with long versions like - `@iconify-prerendered/vue-mdi` to repro.
 If its too hard to get it to work for safari, I wouldn't spend too much time on it.

client/components/BarVersion/BarVersion.scss
```diff
@@ -0,0 +1,46 @@
...
+  line-height: 1;
+  cursor: pointer;
+  height: $bar-width;
+  writing-mode: vertical-rl;
+  text-orientation: sideways;
+  height: 50px;
+  text-align: left;
+  justify-content: flex-start;
+  display: inline-block;
+  margin-top: 5px;
```
pastelsky, line 23: These properties seem to be a no-op —

```css
text-orientation: sideways;
text-align: left; 
justify-content: flex-start; 
display: inline-block;
```

client/components/BarVersion/BarVersion.scss
```diff
@@ -0,0 +1,46 @@
...
+
+  > div {
+    display: flex;
+    text-align: end;
+    justify-content: flex-start;
+    align-items: center;
+    height: 100%;
+    width: 100%;
```
pastelsky, line 31: These properties seem to be a no-op —

```css
text-align: end;
justify-content: flex-start;
height: 100%;
```

pastelsky (changes requested):

darasus:
@pastelsky we can also do truncation with javascript, check last commit

pastelsky (approved):

PR #715, opened 2022-11-03
ts-ify server index

darasus:
- ts-ify server index


index.ts
```diff
@@ -220,14 +226,14 @@ app.prepare().then(() => {
     }
   )
 
-  router.get('/result', async ctx => {
+  router.get('/result', async (ctx: Context) => {
     const packageString = ctx.query.p
-    ctx.redirect(`/package/${packageString.trim()}`)
+    ctx.redirect(`/package/${(packageString as string).trim()}`)
```
pastelsky, line 231: What is the type of `packageString` if we don't force typecast here?

index.ts
```diff
@@ -220,14 +226,14 @@ app.prepare().then(() => {
...
     ctx.status = 301
   })
 
-  router.get('*', async ctx => {
-    const parsedUrl = parse(ctx.req.url, true)
+  router.get('*', async (ctx: Context) => {
+    const parsedUrl = parse(ctx.req.url!, true)
```
pastelsky, line 236: IMO, we should be handling null checks using a library like [ts-invariant](https://www.npmjs.com/package/ts-invariant) instead of forcing a cast to non-null

pastelsky (changes requested):
LGTM, except a few minor comments.

index.ts
darasus: `string | string[] | undefined`

darasus:
@pastelsky I've added validation with invariant, please have a look

index.ts
```diff
@@ -78,7 +77,6 @@ app.prepare().then(() => {
         return /(text|json|javascript|svg)/.test(contentType)
       },
       threshold: 2048,
-      flush: require('zlib').Z_SYNC_FLUSH,
```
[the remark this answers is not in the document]
pastelsky: It does, but perhaps it may not have been configured the correct way before —
https://github.com/koajs/compress

index.ts
```diff
@@ -238,8 +244,7 @@ app.prepare().then(() => {
   })
 
   server.use(router.routes())
-  server.listen(port, err => {
-    if (err) throw err
```
pastelsky, line 242: Why are we ignoring errors now?

index.ts
pastelsky: We should handle non-string types then (use `invariant` if needed for this)

pastelsky (changes requested):

index.ts
darasus: According to koa types there is no error available in arguments

index.ts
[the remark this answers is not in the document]
darasus: Okay this is fixed

pastelsky (approved):

PR #717, opened 2022-11-03
Remove postinstall script

igordanchenko:
This is a proposal to remove `postinstall` script.

I'm not entirely sure why it may be needed for an app, so please feel free to toss this PR if there is a valid reason for `postinstall` script in this project.

**Reasoning**

`postinstall` script often interferes with the dev workflow. Here is a sample scenario: 
1) Imagine you are working on a new feature and your code base is not in a perfect shape so it won't pass `tsc` compilation
2) Now imagine you need to add a new dependency to the project, and you run `yarn add <dependency>` which fails due to `yarn build` failure

cc @darasus

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,7 +99,6 @@
     "prod": "yarn run build && yarn start",
     "test": "jest",
     "test:watch": "jest --watch",
-    "postinstall": "yarn run build",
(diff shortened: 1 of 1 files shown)


pastelsky (approved):

pastelsky:
This was probably done for CI, to execute one command less every time the code was updated. But it makes sense to remove it.

PR #724, opened 2022-11-11
TypeScript migration - scan and scan-results pages

igordanchenko:
Migrate `scan` and `scan-results` pages to TypeScript.

### DIFF
--- a/client/api.ts
+++ b/client/api.ts
@@ -52,7 +52,7 @@ export default class API {
   }
 
   static getInfo(packageString: string) {
-    return API.get(`/api/size?package=${packageString}&record=true`)
+    return API.get<any>(`/api/size?package=${packageString}&record=true`)
   }
 
   static getExports(packageString: string) {
--- a/pages/scan-results/ScanResults.tsx
+++ b/pages/scan-results/ScanResults.tsx
@@ -5,7 +5,6 @@ import cx from 'classnames'
 import ReactFlipMove from 'react-flip-move'
 import PromiseQueue from 'p-queue'
 import queryString from 'query-string'
-import invariant from 'ts-invariant'
 
 import Analytics from '../../client/analytics'
 import Stat from '../../client/components/Stat'
@@ -31,11 +30,12 @@ type Package = {
 } & (
   | { promiseState: 'pending'; result?: undefined }
   | { promiseState: 'rejected'; result?: undefined; error: any }
-  | {
-      promiseState: 'fulfilled'
-      result: Result
-    }
+  | { promiseState: 'fulfilled'; result: Result }
 )
+
+type PackageStateUpdate =
+  | { promiseState: 'fulfilled'; result: Result; version?: string | undefined }
+  | { promiseState: 'rejected'; result: undefined; error: any }
 
 type ResultCardProps = {
   index: number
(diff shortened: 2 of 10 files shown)


client/api.ts
```diff
@@ -52,7 +52,7 @@ export default class API {
   }
 
   static getInfo(packageString: string) {
-    return API.get(`/api/size?package=${packageString}&record=true`)
+    return API.get<any>(`/api/size?package=${packageString}&record=true`)
```
pastelsky, line 55: We should type this `any` — share types between the BE and FE.

pages/scan-results/ScanResults.tsx
```diff
@@ -179,7 +229,16 @@ class ScanResults extends Component {
...
-  updatePackageState(pack, state) {
+  updatePackageState(
+    pack: Package,
+    state:
+      | {
+          promiseState: 'fulfilled'
+          result: Result
+          version?: string | undefined
+        }
+      | { promiseState: 'rejected'; result: undefined; error: any }
```
pastelsky, line 240: Let's move this type definition outside this function definition for clarity

pages/scan-results/ScanResults.tsx
```diff
@@ -100,21 +128,42 @@ class ResultCard extends Component {
...
+      packageString: str,
+      ...parsePackageString(str),
+    }))
+
+    invariant(
+      sortMode === undefined ||
+        sortMode === 'alphabetic' ||
+        sortMode === 'size',
+      'invalid sort mode'
+    )
```
pastelsky, line 166: Might be a better idea to use a default sort mode than throw an error to our users here.

pastelsky (changes requested):

pages/scan-results/ScanResults.tsx
igordanchenko: Done

pages/scan-results/ScanResults.tsx
igordanchenko: Done

client/api.ts
igordanchenko: I completely agree, and we'll get there eventually. 

Ideally this should be typed as `Awaited<ReturnType<typeof getPackageStats>>`, but TS isn't happy about it:


> ./pages/scan-results/ScanResults.tsx:189:31
Type error: Property 'version' does not exist on type '{ size: number; gzip: number; parse: { baseParseTime: number; scriptParseTime: number; } | { baseParseTime?: undefined; scriptParseTime?: undefined; } | null; dependencySizes?: { name: string; approximateSize: number; }[] | undefined; ... 6 more ...; peerDependencies?: string[] | undefined; } | { ...; }'.
>  Property 'version' does not exist on type '{ size: number; gzip: number; parse: { baseParseTime: number; scriptParseTime: number; } | { baseParseTime?: undefined; scriptParseTime?: undefined; } | null; dependencySizes?: { name: string; approximateSize: number; }[] | undefined; ... 6 more ...; peerDependencies?: string[] | undefined; }'.

Do you think you can add clear typings to `package-build-stats` package?

pastelsky: `/api/size` might not return `typeof getPackageStats` directly, it augments the response IIRC. 
Also, `package-build-stats` should have types, but in case you need something specific here, do you mind sending a PR there to add an export? Thanks!

igordanchenko: I see. I guess I underestimated the scope of this PR. Closing.

PR #726, opened 2022-11-17
fix: clear suggestions if the input field is an empty string.

vlmlee:
#### Issue
The autocomplete input will remember suggestions made even after clearing the input field.

#### Solution
Clear the `suggestions` state variable if the input field is an empty string.


pastelsky:
Thanks for raising a fix!

pastelsky (approved):

PR #743, opened 2023-01-11
Typescript migration

lit26:
Add typescript migration for the frontend


pastelsky:
Thanks for the pull request @lit26 — really appreciate the time you took to create this. But it looks like this PR also refactors class components to functional components, and makes other logic changes at a few places. 

Due to this —
1. We would lose git history for existing files since these aren't detected as renames
2. It is difficult to review since the diffs are not very meaningful and there are quite a changes occuring at the same time.

I'd request sharding the changes, and do them one at a time —
1. Just the TS migration.
2. Class -> function components
3. Any other logic changes / fixes

lit26:
Will do.

lit26:
Split steps in other PR.

PR #756, opened 2023-02-27
Ts migration

lit26:
Add typescript migration for the frontend


pages/package/[...packageString]/components/ExportAnalysisSection/ExportAnalysisSection.tsx
```diff
@@ -92,7 +106,7 @@ function ExportList({ exports, totalSize, isLoading }) {
                   size={exportDictionary[letter][0].gzip}
                   totalSize={totalSize}
                   name={exportDictionary[letter][0].name}
-                  path={exportDictionary[letter][0].path}
+                  // path={exportDictionary[letter][0].path}
```
Comment (pastelsky) [inline pages/package/[...packageString]/components/ExportAnalysisSection/ExportAnalysisSection.tsx:109]: Why did we need to comment this out?

pages/package/[...packageString]/components/TreemapSection.tsx
```diff
@@ -156,11 +164,11 @@ class TreemapSection extends Component {
                     className="treemap__percent"
                     style={{
                       fontSize: `${
-                        14 + Math.min(dep.percentShare * 1.2, 25)
+                        14 + Math.min((dep.percentShare || 0) * 1.2, 25)
                       }px`,
                     }}
                   >
-                    {dep.percentShare.toFixed(1)}
+                    {(dep.percentShare || 0).toFixed(1)}
```
Comment (pastelsky) [inline pages/package/[...packageString]/components/TreemapSection.tsx:171]: This looks rather verbose. We should ensure that if `percentShare` is nullable, the right fallback value is set on the API side.

pages/package/[...packageString]/components/ExportAnalysisSection/ExportAnalysisSection.tsx
Comment (lit26) [reply pages/package/[...packageString]/components/ExportAnalysisSection/ExportAnalysisSection.tsx:109]: > Why did we need to comment this out?

I think the child component does not use `path`

PR #777, opened 2023-04-19
[fix] autocomplete empty input request

### LINKED ISSUE #770: Getting 404 for empty input
**Describe the bug**
It looks like it is expected behaviour but I don't think we should allow an empty input call from the frontend.

**To Reproduce**
image
Don't type anything in the search box and press enter.
image
It takes you to `https://bundlephobia.com/package` which gives 404.

**Expected behavior**
I believe we should just do nothing when the input is empty.

Let me know if the suggested fix is good and I can make a relevant PR for the same.

upteran:
### Description
- add check before request send
- add little animation on input error (thought it not form with huge list of inputs, can use some animation)
- add tests and fix parser old test (some of them were fail)

### Related Issue
Closes #770

### Context
I've seen few options to fix this:
- add check to handler in component (with/without animation)
- add `required` attr to autocomplete input [(mean this one)](https://www.w3schools.com/tags/tryit.asp?filename=tryhtml5_input_required), it's easier and native, but in diff browser we'll get different design of tooltips. And I thought in this case we can use more complex solution and get animation and controlled view

Check pls and if animation is not need, I can replace it with `required` attr or another way you prefer )

## Screenshots:
example-shake

### DIFF
--- a/__tests__/utils.test.js
+++ b/__tests__/utils.test.js
@@ -1,10 +1,11 @@
-import { parsePackageString } from '../utils/common.utils'
+const { parsePackageString, isEmpty } = require('../utils/common.utils')
 
 describe('parsePackageString', () => {
   it('handles scoped packages correctly', () => {
     expect(parsePackageString('@babel/core@9.8.0')).toEqual({
       scoped: true,
       name: '@babel/core',
+      scope: 'babel',
       version: '9.8.0',
     })
   })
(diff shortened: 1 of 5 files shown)


pastelsky:
Love the little shake! Thanks.

Yedfree (approved):

PR #1008, opened 2026-07-19
Add curated npm package discovery pages

pastelsky:
## Purpose

Create a small, manually curated set of useful landing pages for package-category and package-comparison searches. These pages give search engines stable URLs and give users a direct path from a broad question to Bundlephobia's package reports.

This intentionally does **not** generate every possible package combination. The routes, copy, packages, related links, and sitemap entries all come from one reviewed content list.

## What changes

- add six focused npm package comparison routes
- add six package category routes with semantic comparison tables and related links
- add comparison/category directory pages and a persistent Compare navigation entry
- replace hydration-time recent searches with stable popular-package links
- replace the small flat sitemap with an index and core, package, and guide shards
- generate guide sitemap URLs from the same curated content source as the pages

## Before / after examples

| Search intent | Before | After |
| --- | --- | --- |
| Compare React and Vue | No dedicated indexable page | `/compare/react-vs-vue` with contextual copy, a package table, and links to both reports |
| Find state-management packages | Search individual packages manually | `/categories/state-management` with curated alternatives and related comparisons |
| Discover available guides | No directory | `/compare` and `/categories` expose every curated page through stable internal links |

The sitemap changes from one small manually maintained file to:

```text
/sitemap.xml
  -> /sitemaps/core.xml
  -> /sitemaps/packages.xml
  -> /sitemaps/guides.xml
       includes /compare/react-vs-vue
       includes /categories/state-management
```

## How package measurements work

The batch helper introduced in this PR uses #1007's in-process `PackageAnalysisService` in `cache-only` mode for every curated package. It cannot initialize the builder or enqueue package builds.

- Exact-version cache hit: return measurements without consulting npm.
- Unversioned package: resolve the current version, then read its existing analysis.
- Cache hit: render minified size, gzip size, dependency count, and module signals in the initial HTML.
- Cache miss or service failure: keep the page available, link to the normal package report, and show the explicit missing-measurement fallback.

## Verification

- `npm run build`
- all 12 curated detail routes return HTTP 200
- invalid detail slugs return HTTP 404
- browser back/forward and deep-link refresh preserve canonical state
- sitemap index and all shards return HTTP 200 and pass `xmllint`
- mobile pages have no horizontal overflow at 390 px
- cache-only no-build behavior is covered by #1007's service tests and production integration checks

## Stack

Stacked on #1007 so this PR contains only curated discovery and sitemap changes.


docs/pr-evidence/seo-curated-discovery/category-mobile.png
pastelsky, line 1: Remove these assets. Also the results say not cached and unkown. Doesn't appear this is working?

pastelsky: Removed all five evidence files and their PR-description links in 20b99a80. You are right that the screenshot did not prove the measurement path was working: it came from a direct local Next server where CACHE_SERVICE_ENDPOINT was unset. npm metadata resolved, but cache measurements could not, which is why the explicit Not cached and Unknown fallback appeared. Production supplies the cache-service endpoint; the public production size endpoint returned HTTP 200 with real size, gzip, and dependency data for the curated package checked during this review. The page deliberately reads existing analyses and never triggers builds, so a genuine cache miss still keeps the page available with the fallback and package-report link. I updated the PR description to make that data contract and the misleading local setup explicit.

Issue #99, opened 2018-10-07
Can we have size badge


MurhafSousli:
## Type

Feature Request

It would be really if we can have a size badge that we can add in the README page, something like:

[![npm](https://img.shields.io/badge/Size-3.2kb-brightgreen.svg)](https://murhafsousli.github.io/ngx-progressbar/)

fabienjuif:
Hi @MurhafSousli you have some pointers here: https://github.com/pastelsky/bundlephobia/issues/91

-> It already exists from tiers apps 👍

styfle:
Also here https://badgen.net/#bundlephobia

MurhafSousli:
@fabienjuif Got it 😄 thanks!

PR #1062, opened 2026-08-05
Add a shared package installation service

pastelsky:
## Summary

- add a singleton installation service alongside the existing main, build, and cache services
- accept npm registry versions, tags, and ranges, resolving them immediately before installation
- deduplicate active exact-version installs with the existing `server/Queue.ts`
- store completed installations in deterministic filesystem directories that survive service restarts
- let every build write entries and Rspack output to its own temporary artifact directory
- configure all three build workers to use the optional service while retaining package-build-stats's local fallback

## Information flow

The public results API already resolves a package for its cache and response metadata, but that is not an installation-service contract. Build callers may submit an exact version, a tag such as `latest`, or a semver range.

The installation service owns the invariant at the natural boundary: immediately before package-manager work, it uses the repository's existing `npm-package-arg`, `semver`, and `pacote` dependencies to turn the requested registry spec and any optional additional packages into canonical `name@exactVersion` strings. Only those exact strings reach the queue and filesystem store. Its flow is:

1. validate that the request is an npm registry package spec
2. normalize an exact version locally, or resolve a tag/range with `pacote.manifest`
3. hash the resolved exact package/version, content-affecting install options, platform, architecture, and Node major
4. attach to an existing active queue job when that key is already installing
5. otherwise reuse the completed deterministic directory or install into a temporary directory and atomically move it into place
6. return the resolved installation path in the same blocking HTTP request

This means mutable requests are resolved afresh, while immutable installations remain safely reusable. Git, URL, and local-path specs are still rejected at this remote-service trust boundary.

There are no leases, release requests, workspace symlinks, Redis locks, PM2 affinity rules, or cross-worker promises.

## Filesystem and restart behavior

- reusable installations live under `/tmp/tmp-build/installations/<hash>`
- `.bundlephobia-installation.json` is the atomic readiness and last-used record
- an installation-service restart reads that record on demand, so no in-memory state is required for reuse
- startup removes incomplete directories; a periodic sweep removes installations older than the 20-minute retention threshold
- the retention threshold is longer than Bundlephobia's ten-minute worker timeout, so an installation is not removed while a valid analysis can still be reading it
- entries and Rspack outputs live under UUID-backed `/tmp/tmp-build/artifacts/...` directories and are removed in `finally`
- completed installation directories contain package-manager files and metadata only—no generated entries or bundles

## Compatibility

- public Bundlephobia endpoints and response shapes are unchanged
- callers are not required to pre-resolve tags or ranges
- results-page behavior remains size, then exports, then export sizes
- package-build-stats remains self-sufficient and installs locally when no service is configured
- a service transport outage falls back locally; installation failures remain authoritative
- this remains a stacked draft on [package-build-stats #99](https://github.com/pastelsky/package-build-stats/pull/99)
- after #99 publishes, both dependency roots must be pinned to that exact package-build-stats release and both lockfiles updated

## Popular-package validation

The exact branches in these two PRs ran stats, exports, and export sizes for all ten exact versions below. The final diagnostics showed ten reusable installations, an empty queue, and zero remaining artifact directories.

| Repository/package | Minified bytes | Gzip bytes | Discovered exports | Export assets | Installations |
|---|---:|---:|---:|---:|---:|
| sindresorhus/ky (`ky@1.10.0`) | 11,482 | 4,087 | 3 | 2 | 1 |
| facebook/react (`react@19.1.1`) | 7,687 | 2,932 | 0 | 0 | 1 |
| ReactiveX/rxjs (`rxjs@7.8.2`) | 63,664 | 17,628 | 173 | 173 | 1 |
| date-fns/date-fns (`date-fns@4.1.0`) | 72,102 | 17,874 | 251 | 250 | 1 |
| ai/nanoid (`nanoid@5.1.5`) | 694 | 463 | 5 | 5 | 1 |
| uuidjs/uuid (`uuid@11.1.0`) | 10,239 | 3,928 | 14 | 14 | 1 |
| pmndrs/zustand (`zustand@5.0.7`) | 852 | 487 | 0 | 0 | 1 |
| chalk/chalk (`chalk@5.4.1`) | 5,606 | 2,433 | 13 | 12 | 1 |
| iamkun/dayjs (`dayjs@1.11.13`) | 7,175 | 3,064 | 0 | 0 | 1 |
| epoberezkin/fast-deep-equal (`fast-deep-equal@3.1.3`) | 891 | 439 | 0 | 0 | 1 |

## Verification

- Prettier on every changed source/configuration file
- ESLint: 0 errors, 5 existing unrelated warnings
- Jest: 15 self-contained suites / 86 tests passed
- package-build-stats: 49 fast + 135 slow tests passed; type/lint/format checks and build passed
- production `yarn build` passed
- live `ky@latest` resolved to `ky@2.0.2`; requesting `ky@2.0.2` reused the identical deterministic directory
- live `ky@^1` resolved independently to `ky@1.14.3`
- two concurrent `p-limit@latest` requests resolved to `p-limit@7.3.1` and returned the same installation path
- live concurrent size and export-size requests succeeded against one installation, with no generated files in that installation and zero leftover artifact directories

The legacy `errors-cache.test.ts` suite still requires a separately running server; without it, its seven HTTP assertions fail with `ECONNREFUSED`. The remaining self-contained tests pass.


server/api/PackageWorkerRegistry.ts
```diff
@@ -0,0 +1,113 @@
...
+  ): Promise<'OK' | null>
+  eval(
+    script: string,
+    numberOfKeys: number,
+    key: string,
+    ...args: Array<string | number>
+  ): Promise<unknown>
+}
+
+async function withinRedisDeadline<TResult>(operation: Promise<TResult>) {
```
pastelsky: Redis is deliberately off the critical path after 250 ms. Affinity improves reuse, but build availability must not depend on Redis being connected or responsive.

server/api/PackageWorkerRegistry.ts
```diff
@@ -0,0 +1,113 @@
...
+  })
+
+  try {
+    return await Promise.race([operation, deadline])
+  } finally {
+    if (timer) clearTimeout(timer)
+  }
+}
+
+/** Keeps closely related package analyses on the same addressable worker. */
```
pastelsky: This is affinity, not a distributed lock: Redis stores only an addressable worker URL. Installation promises and workspace lifecycle remain private to package-build-stats inside that process.

server/api/PackageWorkerRegistry.ts
```diff
@@ -0,0 +1,113 @@
...
+    try {
+      let worker = await withinRedisDeadline(this.redis.get(key))
+      if (worker && !endpoints.includes(worker)) {
+        await this.markUnavailable(packageString, worker)
+        worker = null
+      }
+
+      if (!worker) {
+        await withinRedisDeadline(
+          this.redis.set(key, rankedEndpoints[0], 'PX', AFFINITY_TTL_MS, 'NX')
```
pastelsky: SET NX makes the first concurrent web process establish the shared assignment. Callers then read the chosen value, so differing request arrival order cannot split the package across workers.

server/api/PackageWorkerRegistry.ts
```diff
@@ -0,0 +1,113 @@
...
+      if (!worker || !endpoints.includes(worker)) return rankedEndpoints
+
+      await withinRedisDeadline(
+        this.redis.eval(compareAndRefresh, 1, key, worker, AFFINITY_TTL_MS)
+      )
+      return [
+        worker,
+        ...rankedEndpoints.filter(endpoint => endpoint !== worker),
+      ]
+    } catch (error) {
```
pastelsky: Rendezvous ranking is always computed first and is the complete no-Redis behavior. Redis failure therefore changes only affinity persistence, not whether package analysis can proceed.

server/api/buildServiceEndpoints.ts
```diff
@@ -0,0 +1,30 @@
...
+function hash(value: string) {
+  let result = 2166136261
+  for (const character of value) {
+    result ^= character.charCodeAt(0)
+    result = Math.imul(result, 16777619)
+  }
+  return result >>> 0
+}
+
+export function rankBuildServiceEndpoints(
```
pastelsky: Rendezvous hashing gives every web process the same preferred worker while minimizing package movement when a worker is added or removed. It is safer than array-index modulo routing during topology changes.

server/api/BuildService.ts
```diff
@@ -60,40 +61,60 @@ export default class BuildService {
...
-            try {
-              const response = await axios.get(
-                `${process.env.BUILD_SERVICE_ENDPOINT}${
-                  operation.endpoint
-                }?p=${encodeURIComponent(packageString)}`,
-                {
-                  signal,
-                  maxContentLength: MAX_BUILD_SERVICE_RESPONSE_BYTES,
+          try {
+            const buildServiceEndpoints = await packageWorkerRegistry.rank(
```
pastelsky: All three queue operation types call the same package-based ranker. Queue deduplication still handles identical operations; worker affinity enables package-build-stats to reuse installation state across different operations.

server/api/BuildService.ts
```diff
@@ -60,40 +61,60 @@ export default class BuildService {
...
+                      signal,
+                      maxContentLength: MAX_BUILD_SERVICE_RESPONSE_BYTES,
+                    }
+                  )
+                  return response.data
+                } catch (error) {
+                  if (axios.isCancel(error)) {
+                    throw new JobCancelledError()
+                  }
+                  if (
```
pastelsky: Fallback is restricted to transport failures with no HTTP response. A real build error remains authoritative and is not retried on another worker, which avoids duplicate expensive builds and misleading error masking.

build-service/index.js
```diff
@@ -2,15 +2,32 @@ import 'dotenv-defaults/config.js'
...
   getPackageExportSizes,
+  getPackageExports,
   eventQueue,
 } from 'package-build-stats'
 import Amplitude from '@amplitude/node'
 import serializeError from './serializeError.js'
 
 const fastify = Fastify()
 
+async function analyzePackage(req, res, packageString, analyze) {
```
pastelsky, line 14: Request and response-close signals are translated into one AbortSignal for package-build-stats. The package coordinator detaches this subscriber immediately and aborts shared installation work only if no subscribers remain.

process.yml
```diff
@@ -26,14 +26,18 @@ apps:
     node_args:
       - --require=./scripts/memory-diagnostics.cjs
     instances: 3
```
pastelsky, line 28: Stable incremented ports make each process addressable. PM2 cluster load balancing cannot preserve process-local installation affinity because repeated requests to one shared port may land on different workers.

pastelsky (commented):
Architecture notes explaining the routing, optional Redis boundary, failure behavior, and stable worker topology.

installation-service/index.mjs
```diff
@@ -0,0 +1,107 @@
...
+    networkConcurrency: value.networkConcurrency,
+    additionalPackages: value.additionalPackages,
+    installTimeout: value.installTimeout,
+    debug: value.debug,
+    // REST callers may install registry packages, never arbitrary local paths.
+    isLocal: false,
+  }
+}
+
+function isRegistryPackage(packageString) {
```
pastelsky: npm-package-arg is used here as the canonical registry-spec parser. Rejecting file, Git, and remote URL specs keeps the internal service from turning its package input into arbitrary local-path access.

installation-service/InstallationStore.cjs
```diff
@@ -0,0 +1,198 @@
...
+    this.installationApi = installationApi
+    this.idleMs = idleMs
+    this.leaseMs = leaseMs
+    this.onError = onError
+    this.queue = new InstallQueue(concurrency)
+    this.installations = new Map()
+    this.leases = new Map()
+  }
+
+  key(packageString, options) {
```
pastelsky: The key includes every option that can change installed contents or package-manager behavior, plus the host runtime ABI dimensions. Build-only options are excluded so stats and export analyses converge on the same installation.

installation-service/InstallationStore.cjs
```diff
@@ -0,0 +1,198 @@
...
+
+    return this.queue.run(key, async () => {
+      const concurrentlyCached = this.installations.get(key)
+      if (concurrentlyCached) return concurrentlyCached
+
+      const installation = await this.installationApi.installPackage(
+        packageString,
+        options
+      )
+      const canonicalKey = this.key(
```
pastelsky: The canonical exact-version alias matters for the result flow: an initial tag or range may be followed by requests using the resolved name@version. Both keys point at one base installation and are removed together.

installation-service/index.mjs
```diff
@@ -0,0 +1,107 @@
...
+  try {
+    return ['tag', 'version', 'range'].includes(npa(packageString).type)
+  } catch {
+    return false
+  }
+}
+
+async function installPackageWithVersion(packageString, options) {
```
pastelsky, line 38: The service reads the installed manifest only to register the canonical exact-version alias. Package-manager execution and workspace construction stay owned by package-build-stats; Bundlephobia owns only coordination policy.

installation-service/InstallationStore.cjs
```diff
@@ -0,0 +1,198 @@
...
+
+class InstallQueue {
+  constructor(concurrency) {
+    this.concurrency = concurrency
+    this.active = 0
+    this.pending = []
+    this.inFlight = new Map()
+  }
+
+  run(key, task) {
```
pastelsky: This is the installation-specific single-flight boundary. Requests with the same key receive the same promise, while the small queue also caps unrelated package-manager processes; build work is intentionally outside this queue.

process.yml
```diff
@@ -34,10 +34,27 @@ apps:
     env:
       NODE_ENV: production
       DEBUG: 'bp:*'
+      INSTALLATION_SERVICE_ENDPOINT: http://127.0.0.1:7003
       MEMORY_DIAGNOSTICS_SERVICE: build-service
       MEMORY_DIAGNOSTICS_RSS_THRESHOLD: 850M
       MEMORY_DIAGNOSTICS_DIR: /var/cache/bundlephobia/diagnostics
 
+  - script: ./installation-service/index.mjs
```
pastelsky, line 42: One installation-service process is authoritative for this host, while the three build workers remain an ordinary PM2 cluster. No stable worker ports, sticky routing, Redis, or distributed filesystem lock is required.

installation-service/InstallationStore.cjs
```diff
@@ -0,0 +1,198 @@
...
+      workspace = await this.installationApi.createPackageWorkspace(
+        entry.installation
+      )
+    } catch (error) {
+      entry.leases -= 1
+      this.scheduleCleanup(entry)
+      throw error
+    }
+
+    const id = randomUUID()
```
pastelsky: Lease expiry is the backstop for a crashed worker or lost DELETE request. Normal clients release immediately; this timer bounds disk retention when the client disappears after acquire.

installation-service/InstallationStore.cjs
```diff
@@ -0,0 +1,198 @@
...
+          this.installations.delete(key)
+      }
+      void this.installationApi
+        .disposePackage(entry.installation)
+        .catch(this.onError)
+    }, this.idleMs)
+    entry.cleanupTimer.unref()
+  }
+
+  async acquire(packageString, options = {}) {
```
pastelsky: Acquire never hands out the cached base directory. It increments a lease first and asks package-build-stats for a separate writable workspace, which allows builds on different PM2 workers to run concurrently without output races.

installation-service/InstallationStore.cjs
```diff
@@ -0,0 +1,198 @@
...
+    })
+  }
+
+  cancelCleanup(entry) {
+    if (!entry.cleanupTimer) return
+    clearTimeout(entry.cleanupTimer)
+    entry.cleanupTimer = undefined
+  }
+
+  scheduleCleanup(entry) {
```
pastelsky: Cleanup starts only after the final lease and waits through the short results-page gap. A new acquire cancels this timer, allowing size, exports, and export-size requests to reuse the base without retaining it indefinitely.

pastelsky:
Architecture update: commit c8f48766 completely removes the Redis/worker-affinity approach from the final diff. The active design is the singleton installation service described in the PR body. Nine inline self-review comments on c8f48766 explain queue single-flight behavior, keying, canonical aliases, workspace isolation, lease cleanup, input policy, and PM2 topology.

installation-service/createInstallQueue.cjs
```diff
@@ -0,0 +1,22 @@
...
+})
+const Queue = require('../server/Queue').default
+
+/** Reuses the main-service queue for package/options keyed installation tasks. */
+module.exports = function createInstallQueue(concurrency) {
+  const queue = new Queue({ concurrency, aging: false })
+  queue.addExecutor('install-package', ({ task }) => task())
+
+  return {
+    run: (key, task) =>
```
pastelsky, line 16: Queue.process already deduplicates by id and job type, promotes attached callers to listeners, and enforces concurrency. The adapter contributes only the installation job type and exposes existing diagnostics; it does not reimplement queue behavior.

installation-service/createInstallQueue.cjs
```diff
@@ -0,0 +1,22 @@
+const { register } = require('esbuild-register/dist/node')
+const tsconfig = require('../tsconfig.server.json')
+
+register({
+  tsconfigRaw: tsconfig,
+  target: tsconfig.compilerOptions.target,
+})
+const Queue = require('../server/Queue').default
```
pastelsky, line 8: This adapter loads the same TypeScript Queue used by the main service. Keeping the service entrypoint in JavaScript avoids making the stacked web build resolve the not-yet-published package-build-stats subpath at typecheck time.

pastelsky:
Queue reuse follow-up: f323518 removes the custom InstallQueue completely. The installation service now submits package/options keyed tasks through the existing server/Queue.ts; a live three-endpoint concurrent run showed one running job, two attached listeners, one installation lifecycle, three HTTP 200 responses, then zero queued jobs and zero leases.

installation-service/InstallationStore.cjs
```diff
@@ -0,0 +1,155 @@
...
+      workspace = await this.installationApi.createPackageWorkspace(
+        entry.installation
+      )
+    } catch (error) {
+      entry.leases -= 1
+      this.scheduleCleanup(entry)
+      throw error
+    }
+
+    const id = randomUUID()
```
pastelsky: The lease UUID is separate from the Queue job ID because installation work may already be complete when an analysis begins. Expiry reclaims workspaces from crashed workers without cancelling or mutating the cached base.

installation-service/InstallationStore.cjs
```diff
@@ -0,0 +1,155 @@
...
+      process.versions.node.split('.')[0],
+    ])
+  }
+
+  async getInstallation(packageString, options) {
+    const key = this.key(packageString, options)
+    const cached = this.installations.get(key)
+    if (cached) return cached
+
+    return this.queue.run(key, async () => {
```
pastelsky: This call is now the sole concurrency and in-flight deduplication mechanism. The second cache check remains necessary because a queued job may start after an earlier job populated the short-lived installation store.

installation-service/InstallationStore.cjs
```diff
@@ -0,0 +1,155 @@
...
+    this.installationApi = installationApi
+    this.idleMs = idleMs
+    this.leaseMs = leaseMs
+    this.onError = onError
+    this.queue = queue
+    this.installations = new Map()
+    this.leases = new Map()
+  }
+
+  key(packageString, options) {
```
pastelsky: Only installation-affecting options and runtime dimensions participate in this ID. That lets the existing Queue attach stats, exports, and export-size callers while keeping genuinely different installs separate.

installation-service/InstallationStore.cjs
```diff
@@ -0,0 +1,155 @@
...
+          this.installations.delete(key)
+      }
+      void this.installationApi
+        .disposePackage(entry.installation)
+        .catch(this.onError)
+    }, this.idleMs)
+    entry.cleanupTimer.unref()
+  }
+
+  async acquire(packageString, options = {}) {
```
pastelsky: A Queue result is only the immutable base installation. Every subscriber then creates its own leased workspace, keeping build outputs isolated even when the shared Queue releases multiple listeners together.

installation-service/InstallationStore.cjs
```diff
@@ -0,0 +1,155 @@
...
+
+    return this.queue.run(key, async () => {
+      const concurrentlyCached = this.installations.get(key)
+      if (concurrentlyCached) return concurrentlyCached
+
+      const installation = await this.installationApi.installPackage(
+        packageString,
+        options
+      )
+      const canonicalKey = this.key(
```
pastelsky: Registering the resolved exact version as an alias lets a tag or range request feed a later canonical result-page request. Both map keys point to the same entry and cleanup removes only keys still owned by that entry.

installation-service/InstallationStore.cjs
```diff
@@ -0,0 +1,155 @@
...
+    })
+  }
+
+  cancelCleanup(entry) {
+    if (!entry.cleanupTimer) return
+    clearTimeout(entry.cleanupTimer)
+    entry.cleanupTimer = undefined
+  }
+
+  scheduleCleanup(entry) {
```
pastelsky: This idle timer is installation retention policy, not queue behavior. Queue owns active job lifetime; the store keeps a completed base just long enough for the next distinct operation to reuse it.

pastelsky:
Architecture simplification in c0f3df3b: installations are now deterministic, immutable filesystem cache entries keyed by exact package version and options. The service returns the path in one blocking POST; leases, DELETE releases, symlink workspaces, and in-memory installation ownership are removed. Readiness/last-use metadata makes reuse and cleanup restart-safe, while package-build-stats owns UUID-backed temporary artifacts.

installation-service/InstallationStore.cjs
```diff
@@ -0,0 +1,180 @@
...
+  ) {
+    this.installationApi = installationApi
+    this.queue = queue
+    this.rootPath = rootPath
+    this.retentionMs = retentionMs
+    this.onError = onError
+    this.sweepTimer = undefined
+  }
+
+  key(packageString, options) {
```
pastelsky: The cache identity starts with the already resolved exact package version and includes every installation-affecting option plus runtime compatibility fields. Tags never reach this layer.

installation-service/InstallationStore.cjs
```diff
@@ -0,0 +1,180 @@
...
+      options.networkConcurrency,
+      options.additionalPackages,
+      options.installTimeout,
+      process.platform,
+      process.arch,
+      process.versions.node.split('.')[0],
+    ])
+  }
+
+  directory(key) {
```
pastelsky, line 36: A deterministic directory makes the filesystem authoritative across service restarts. The truncated SHA-256 is only a path name; the full unhashed key is retained and checked in metadata.

installation-service/InstallationStore.cjs
```diff
@@ -0,0 +1,180 @@
...
+    const hash = createHash('sha256').update(key).digest('hex').slice(0, 24)
+    return path.join(this.rootPath, hash)
+  }
+
+  async read(directory) {
```
pastelsky, line 41: The metadata file is both the readiness marker and restart record. Paths are constrained to the expected cache directory before a restarted process returns them to a build worker.

installation-service/InstallationStore.cjs
```diff
@@ -0,0 +1,180 @@
...
+    const metadataPath = path.join(
+      entry.installation.installPath,
+      METADATA_FILE
+    )
+    const temporaryPath = `${metadataPath}.tmp`
+    await fs.writeFile(temporaryPath, JSON.stringify(entry))
+    await fs.rename(temporaryPath, metadataPath)
+  }
+
+  async install(key, packageString, options) {
```
pastelsky: Package managers install into their normal temporary directory first. A same-filesystem rename publishes the completed tree atomically, so a crash cannot expose a half-installed cache entry.

installation-service/InstallationStore.cjs
```diff
@@ -0,0 +1,180 @@
...
+        installPath: directory,
+        packagePath: path.join(directory, relativePackagePath),
+      }
+    } catch (error) {
+      await this.installationApi.disposePackage(installed)
+      throw error
+    }
+  }
+
+  async get(packageString, options = {}) {
```
pastelsky: The existing Queue coordinates only active creation for this key. Completed reuse is a direct filesystem read, so no promise, lease, or in-memory cache must survive a restart.

installation-service/InstallationStore.cjs
```diff
@@ -0,0 +1,180 @@
...
+  async entries() {
+    const directories = await fs.readdir(this.rootPath, { withFileTypes: true })
+    return Promise.all(
+      directories
+        .filter(entry => entry.isDirectory())
+        .map(entry => this.read(path.join(this.rootPath, entry.name)))
+    )
+  }
+
+  async sweep() {
```
pastelsky, line 133: Incomplete entries have no valid metadata and are removed on startup. Completed entries are retained beyond the ten-minute worker timeout before age-based eviction can remove them.

installation-service/index.mjs
```diff
@@ -0,0 +1,114 @@
...
+    networkConcurrency: value.networkConcurrency,
+    additionalPackages: value.additionalPackages,
+    installTimeout: value.installTimeout,
+    debug: value.debug,
+    // REST callers may install registry packages, never arbitrary local paths.
+    isLocal: false,
+  }
+}
+
+function isExactRegistryPackage(packageString) {
```
pastelsky: This second boundary check guarantees cache keys are exact versions even if a caller bypasses Bundlephobia's public resolver and calls the build service directly.

installation-service/index.mjs
```diff
@@ -0,0 +1,114 @@
...
+})
+
+fastify.get('/diagnostics', async () => store.diagnostics())
+fastify.addHook('onClose', async () => store.close())
+
+const port = positiveInteger(process.env.PORT, 7003)
+await fastify.listen({ host: '127.0.0.1', port })
+console.log(`installation service listening on ${port}`)
+
+async function shutdown() {
```
pastelsky, line 121: Graceful PM2 stops close Fastify and cancel the sweeper, but intentionally retain completed filesystem entries so the next process can reuse them immediately.

pastelsky (commented):
Self-review notes for the simplified exact-version filesystem cache.

installation-service/resolveRegistryPackageSpec.cjs
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  if (!spec.registry || !['tag', 'range', 'version'].includes(spec.type)) {
+    throw new UnsupportedRegistryPackageSpecError(
+      'packageString must be an npm registry package spec'
+    )
+  }
+
+  if (spec.type === 'version') {
+    return `${spec.name}@${semver.clean(spec.fetchSpec)}`
```
pastelsky, line 26: Exact versions do not need a registry round trip. semver.clean canonicalizes equivalent spellings such as v1.2.3 and =1.2.3 so they cannot create duplicate installation keys.

installation-service/index.mjs
```diff
@@ -0,0 +1,118 @@
...
+    return reply.code(400).send({
+      name: 'InstallError',
+      originalError: 'packageString must be an npm registry package spec',
+    })
+  }
+
+  try {
+    // Resolve mutable tags/ranges immediately before queueing so the cache and
+    // filesystem are always keyed by the exact version that will be installed.
+    const exactPackageString = await resolveRegistryPackageSpec(
```
pastelsky: Bundlephobia callers may still send latest or a range. Resolution happens here, immediately before queueing, because this service is the component that can authoritatively bind the request to the version it installs.

installation-service/resolveRegistryPackageSpec.cjs
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  const resolved = await manifest(packageString, { fullMetadata: false })
```
pastelsky, line 29: Tags and ranges are intentionally resolved on every request at the installation boundary. pacote supplies npm-compatible selection semantics, and only its resolved name@version is allowed into persistent reuse.

installation-service/InstallationStore.cjs
```diff
@@ -0,0 +1,179 @@
...
+  ) {
+    this.installationApi = installationApi
+    this.queue = queue
+    this.rootPath = rootPath
+    this.retentionMs = retentionMs
+    this.onError = onError
+    this.sweepTimer = undefined
+  }
+
+  key(exactPackageString, options) {
```
pastelsky, line 25: The store API names this exactPackageString deliberately: mutable specs have already been resolved, so both in-flight deduplication and restart-safe filesystem reuse share one immutable identity.

installation-service/index.mjs
```diff
@@ -0,0 +1,118 @@
...
+    return await store.get(exactPackageString, installOptions(body.options))
+  } catch (error) {
+    if (error instanceof UnsupportedRegistryPackageSpecError) {
+      return reply.code(400).send({
+        name: 'InstallError',
+        originalError: error.message,
+      })
+    }
+    const serialized = serializeError(error)
+    const packageNotFound = ['E404', 'ETARGET'].includes(error?.code)
```
pastelsky, line 107: Registry misses remain authoritative package errors rather than service-availability failures. That distinction prevents package-build-stats from falling back locally and starting a duplicate install for an invalid spec.

installation-service/resolveRegistryPackageSpec.cjs
```diff
@@ -0,0 +1,42 @@
...
+  manifest = pacote.manifest
+) {
+  let spec
+  try {
+    spec = npa(packageString)
+  } catch (error) {
+    throw new UnsupportedRegistryPackageSpecError(error.message)
+  }
+
+  if (!spec.registry || !['tag', 'range', 'version'].includes(spec.type)) {
```
pastelsky, line 19: This is the remote-service trust boundary, not a restriction on normal npm usage: versions, tags, and ranges are accepted, while Git, URL, and local-path specs remain unavailable to an internal HTTP caller.

installation-service/index.mjs
```diff
@@ -0,0 +1,126 @@
...
+      name: 'InstallError',
+      originalError: 'packageString must be an npm registry package spec',
+    })
+  }
+
+  try {
+    // Resolve mutable tags/ranges immediately before queueing so the cache and
+    // filesystem are always keyed by the exact version that will be installed.
+    const options = installOptions(body.options)
+    const [exactPackageString, ...exactAdditionalPackages] = await Promise.all(
```
pastelsky, line 90: The primary spec and every additional package are resolved together immediately before queueing. This closes the same stale-cache and local-path hole for optional installs without making package-build-stats aware of registry policy.

pastelsky:
Architecture update in c0b46c20: the installation service now returns a `subscriptionId`, tracks active subscribers per canonical installation, accepts `DELETE /installations/:id`, and retains an installation for five minutes after the last unsubscribe. A persisted ten-minute lease protects in-flight work across service restarts. package-build-stats releases subscriptions in `finally`; per-analysis artifacts remain isolated and temporary.

pastelsky:
Follow-up fix in 2308d2d2: the existing keyed mutex remains intentionally small and distinct from the install queue—the queue coalesces installs, while the mutex gives each caller its own subscription. Expired in-memory subscriptions are now swept too, so a client that dies before DELETE cannot retain an installation forever. Added coverage for abandoned subscriptions.

PR #1075, opened 2026-08-07
feat: add language-owned storage adapters

pastelsky:
## Summary

- define capability-oriented storage interfaces for package analysis results, export-size results, package history, and recent searches
- add separate JavaScript, Java, and Kotlin storage adapters
- make JavaScript delegate to the existing cache-service and Firebase utilities without changing their roots, payloads, fallback behavior, or version ordering
- leave Java and Kotlin adapters without storage capabilities until their actual backends are implemented
- route current JavaScript storage reads and writes through the JavaScript adapter

## Design boundary

Each language adapter owns its storage implementation. There is no shared table layout, Firebase namespace model, sharding policy, schema/profile key scheme, or assumption that languages use the same database. A future adapter can use new SQL tables, Firebase roots, files, or a remote service without changing JavaScript storage.

## Backward compatibility

The existing `Cache`, cache-service middleware, `FirebaseUtils`, Firebase roots, semver history behavior, and public API contracts are unchanged. The JavaScript adapter only delegates to those existing implementations.

## Verification

- 16 unit suites / 105 tests passed (excluding the pre-existing live `errors-cache.test.ts`)
- `yarn lint` passed (5 pre-existing warnings)
- `yarn build` passed
- Playwright production E2E: 8/8 passed with no API mocks, including a real `is-number@7.0.0` analysis
- `git diff --check` passed


server/middlewares/generateImg.middleware.ts
```diff
@@ -51,10 +49,11 @@ const generateImgMiddleware: Middleware = async ctx => {
       resolvedVersion = version
     }
 
-    const result = await cache.getPackageSize<StatsImageResult>({
-      name,
-      version: resolvedVersion,
-    })
+    const result =
+      await javascriptStorage.packageAnalysis.get<StatsImageResult>({
```
pastelsky: Ideally all of this also should not be assumed always to be JS

server/middlewares/results/build.middleware.ts
```diff
@@ -97,11 +95,11 @@ const buildMiddleware: Middleware = async ctx => {
   )
 
   if (record === 'true') {
-    firebaseUtils.setRecentSearch(name, { name, version })
+    javascriptStorage.recentSearches.record(name, { name, version })
```
pastelsky: Ideally all of this also should not be assumed always to be JS. Do we plan on switch casing somewhere above?

pastelsky: Fixed in aa6e4789. There is no switch: createAnalysisContextMiddleware carries the route language, and resolution, cache lookup, build recording, force writes, exports, and export-size middleware select the adapter with getLanguageStorageAdapter(ctx.state.analysis.language). Existing legacy routes still default to JavaScript.

server/middlewares/generateImg.middleware.ts
pastelsky: Fixed in aa6e4789. The stats-image route now establishes an analysis context, and the middleware uses ctx.state.analysis.language for both package resolution and getLanguageStorageAdapter(language). The legacy route supplies JavaScript at the route boundary; the middleware itself no longer assumes it.

PR #1081, opened 2026-08-10
Replace Umami with Amplitude Analytics

pastelsky:
## Summary
- use the maintained `@amplitude/analytics-browser` SDK with a lazy browser-only import
- relay analytics through a neutral same-origin `/_events` endpoint
- add an HAProxy edge backend that forwards `/_events` to Amplitude’s HTTP API over verified TLS
- reject non-POST event requests and apply the existing client rate limit

## Verification
- focused analytics tests: passed (5/5)
- production build: passed previously
- HAProxy config: validated by the live HAProxy parser via stdin
- live `/etc/haproxy/haproxy.cfg` matched the repository before this change
- full Jest: 110 passed, 4 unrelated existing cache/build fixture failures
- prior E2E CI failure was an unrelated package-analysis timeout; a fresh CI run is expected after this push

After deployment, verify a browser event reaches `/_events` and appears in Amplitude.


client/__tests__/analytics.test.ts
```diff
@@ -1,43 +1,66 @@
...
   })
 
   afterEach(() => {
     Reflect.deleteProperty(globalThis, 'window')
   })
 
-  it('sends page context through Umami without duplicating pageviews', () => {
+  it('initializes Amplitude only once in the browser', async () => {
+    initializeAmplitude()
+    initializeAmplitude()
```
pastelsky: Duplicate.

PR #1089, opened 2026-08-10
feat: add trends cache service endpoints

pastelsky:
## Summary

- add typed cache names and keys for trend data
- expose bounded in-memory LRU/TTL endpoints for downloads, GitHub history, releases, and size history
- add matching methods to the server cache client

This is a focused foundation slice; trend data-source services will follow separately.

## Validation

- `yarn typecheck`
- direct TypeScript check of the new cache middleware
- lint (existing repository warnings only)

### DIFF
--- a/cache-service/index.ts
+++ b/cache-service/index.ts
@@ -30,15 +30,15 @@ fastify.post('/package-cache', postPackageSizeMiddlware)
 fastify.get('/exports-cache', getExportsSizeMiddlware)
 fastify.post('/exports-cache', postExportsSizeMiddleware)
 
-const trendsCacheNames: TrendsCacheName[] = [
-  'downloads',
-  'github-history',
-  'releases',
-  'size-history',
-]
+const trendsCacheConfig: Record<TrendsCacheName, number> = {
+  downloads: 24 * 60 * 60 * 1000,
+  'github-history': 24 * 60 * 60 * 1000,
+  releases: 24 * 60 * 60 * 1000,
+  'size-history': 24 * 60 * 60 * 1000,
+}
 
-for (const cacheName of trendsCacheNames) {
-  const middleware = createTrendsCacheMiddleware()
+for (const [cacheName, ttlMs] of Object.entries(trendsCacheConfig)) {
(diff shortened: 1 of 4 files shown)


pastelsky:
Updated per feedback:

- Removed caller-supplied `ttlMs` from the trends-cache API; each cache namespace now owns its TTL in the cache service.
- Replaced the generic public `(cacheName, key, ...)` client API with named operations (`getDownloads`, `setDownloads`, `getGithubHistory`, etc.).
- Kept the cache namespace-to-route mapping internal to the cache service.
- The wire request now carries only the opaque cache key and result.

pastelsky:
Follow-up to the previous update: the cache API no longer accepts a generic key either. Each endpoint now has an explicit domain request:

- downloads: `packageName` + `range`
- github-history: `repository` + `source` + optional `range`
- releases/size-history: `packageName`

The cache service derives its internal LRU key from those fields, while TTL and namespace routing remain service-owned.

pastelsky:
Data-source comparison: the existing `/api/package-history` reads Firebase build snapshots (size/gzip per version) and uses Algolia only to determine the recent version set. The trends releases source reads npm packument `time` data because it needs publish dates and release classification across versions.

The separate `size-history` cache was redundant, so it has been removed. The future size-series service should reuse package history with a larger limit and join those snapshots to the release publish-date map.

pastelsky:
Why releases still use the npm registry: the existing Algolia response exposes the current version and a version-name map; package history exposes Firebase build snapshots keyed by version (`size`/`gzip`). Neither contract provides authoritative publish timestamps.

We can reuse Algolia/package history for version selection and size data, but the trends release timeline still needs npm packument `time` metadata (or another timestamp-bearing source). Replacing it with Algolia would lose the date axis and major/minor release placement.

pastelsky:
Adjusted cache policy:

- downloads: 6h
- releases: 6h
- GitHub history: 12h
- GitHub stars history: 12h
- live GitHub snapshot: 1h

GitHub history, stars, and snapshot now have separate endpoints, cache namespaces, key shapes, and client methods. A fresh live snapshot will not be hidden behind a historical-series cache.

PR #1108, opened 2026-09-12
fix: make general-purpose-date-time reachable by the category classifier

leemr:
## What a visitor sees today

[`spacetime`](https://bundlephobia.com/package/spacetime), [`@js-joda/core`](https://bundlephobia.com/package/@js-joda/core), [`date-and-time`](https://bundlephobia.com/package/date-and-time) and [`date-fns-tz`](https://bundlephobia.com/package/date-fns-tz) are all general-purpose date libraries, and every one of them is offered `chrono-node`, `its-a-date` and `parse-messy-time` as its similar packages — natural-language date parsers. None of them parses natural language. Each link shows it.

They are not in any `similar` array, so they reach the classifier by scoring, and the scoring path cannot return `general-purpose-date-time` for anything.

For disclosure: this turned up while looking at why [`daymath`](https://bundlephobia.com/package/daymath), a date package I maintain, was offered the same three parsers. It is also uncurated, and this change does not add it, or anything else, to a `similar` list. The four named above are unrelated to me and reproduce it on their own.

## Why

`getCategory` picks the winner with a strict comparison:

```js
if (score > maxScoreCategory.score) {
  maxScoreCategory = { label, score }
}
```

A tie therefore keeps whichever category is declared first. `general-purpose-date-time` has tags that are a **strict subset** of `date-nlp`'s, at identical weights, and `date-nlp` is declared first. It can tie but never win.

`react-animation` has the same shape against `general-purpose-animation`. It is also the only one of the eleven `react-*` categories with no tag containing `react`.

## Why it has been invisible

The five packages curated under `general-purpose-date-time` never hit this path. `getInCategoryMap` runs first and returns their declared category at score `999`, so `date-fns`, `dayjs`, `luxon`, `moment` and `@formkit/tempo` all look correct on the site. Only uncurated packages fall through to scoring, and those are exactly the packages nobody checks.

`similarPackages.middleware.ts` had an `async function test()` asserting that every curated package classifies into the category declaring it. It sits behind `void test` and never runs — but it would not have caught this even if it did, because it calls `getCategory`, which short-circuits for precisely the packages it iterates. The check was unreachable twice over.

## The fix

Weight each category's distinctive tags above the generic ones it shares with an earlier category.

For `date-nlp`: `nlp` and `human` rise to `MAX`, `natural language` stays `HIGH`, and `date`, `time` and `format` drop below the weights `general-purpose-date-time` gives them. `parse` and `parser` are unchanged at `MID` in both — they did not need to move, because the three that did are enough to break the tie.

For `react-animation`: a `react` tag at `Weight.NORMAL`, matching `react-autocomplete`, `react-head-meta`, `react-i18n` and `react-form`.

Scoring the packages named in a `similar` array against their published npm metadata, seven change category and six of those become correct:

```
@formkit/tempo   date-nlp              -> general-purpose-date-time   now correct
moment           date-nlp              -> general-purpose-date-time   now correct
luxon            date-nlp              -> general-purpose-date-time   now correct
dayjs            date-nlp              -> general-purpose-date-time   now correct
date-fns         date-nlp              -> general-purpose-date-time   now correct
framer-motion    react-native-charting -> react-animation             now correct
globalize        date-nlp              -> general-purpose-date-time   wrong before and after
```

`globalize` is declared under `i18n` and matches neither category either way, so it is untouched by this change. Nothing regresses.

Two of the 224 distinct names, `suffix-tree` and `bin-tree`, are not on npm and could not be scored. Three names appear in two categories each — `js-cookie`, `universal-cookie` and `axios` — so an aggregate "how many agree" figure depends on which of their two categories is treated as declared. That is why the evidence above is the named moves rather than a total.

## The test

`__tests__/similar-packages.test.ts` replaces the dead helper and runs under `yarn test`. To make the scoring testable without an Algolia round trip, it is extracted from `getCategory` as an exported `classifyPackage`. Behavior is unchanged, argument defaults included, and `getCategory` keeps its `getInCategoryMap` short circuit before calling it. Calling `classifyPackage` directly is what lets the test reach the scoring path at all.

Two parts. The first asserts both date categories from published npm metadata. The second asserts that no category is unreachable: for each category, no earlier-declared category may carry every one of its tag tokens at the same or a higher weight.

That second check compares **tokens**, not raw tag strings, because that is what scoring compares. `animation` and `animations` are one token to the scorer, since both stem to `anim`, and two different strings to a naive check.

Verified by restoring `react-animation`'s pre-change tags and renaming `animation` to `animations`. The token check fails and names `react-animation can never outscore general-purpose-animation`; a raw-string comparison reports nothing. The rename on its own is not enough against the fixed data, because the `react` tag added here already makes that category reachable.

`MIN_CUTOFF_SCORE` and the `classifier:test` logger are removed with the helper, their only reader. The constant is not applied on the live path, and `date-fns` and `luxon` both score below it, so asserting it would have asserted a rule the current data does not satisfy.

## Found and deliberately not fixed

**A multi-word tag becomes independent tokens, each keeping the tag's whole weight.** `{ tag: 'react native', weight: Weight.MAX }` yields `react` and `nativ` at 15 each, so any package with a `react` keyword scores 15 in `react-native-charting`. `natural language` behaves the same way for `language`. Changing that would alter scoring in every category, so it is out of scope here.

**`react-spring` remains misclassified**, into `react-native-charting`, because of its `react-native` keyword and the rule above. This change does not fix it. Its npm `description` is also a raw HTML block, which turns out to be irrelevant: it classifies identically with that description decoded, stripped, or empty.

## Verification

```
yarn lint
yarn typecheck
yarn test:recommendations
yarn build
yarn jest __tests__/similar-packages.test.ts
```

All pass, on Node 24.13.0 from `.nvmrc` and on Node 26 from the CI matrix.

Merged the latest `bundlephobia` in, so this now sits on the anti-slop lint baseline.  `yarn lint`, `yarn typecheck`, `yarn format:check`, and `yarn jest __tests__/similar-packages.test.ts` all pass against it, and the test file picked up a `satisfies`, two blank lines, and `moment`'s current npm keywords.

`yarn jest` as a whole has one failing suite, `__tests__/errors-cache.test.ts`. It needs a server on `127.0.0.1:5000`; with `yarn start` running, four of its seven tests pass and three fail on a `cache-control` assertion expecting `SIZE_API_HAS_VERSION` where the server returns `SIZE_API_DEFAULT`. No CI job runs that suite, and none of it touches similar packages.


leemr:
Heads up, the `bundlephobia` rename didn't reach `.github`: `check-recommendation-changes.mjs:13` and `package-recommendations.yml:9` should point to `server/middlewares/similar-packages/similar-packages.fixtures.ts`.

PR #1109, opened 2026-09-13
Enable first anti-slop lint batch

pastelsky:
## What this fixes

Enables the first anti-slop lint batch as hard errors:

- `no-constant-binary-expression`
- `no-unreachable`
- `no-useless-catch`
- `no-useless-assignment`
- `no-useless-escape`
- `no-unused-expressions`
- `no-unused-labels`
- `no-unsafe-finally`
- `no-unsafe-optional-chaining`

Existing violations were fixed across unused initial assignments and dead control-flow patterns. The cleanup also makes the pruning reason and progress state explicitly derived before use. The originally proposed `no-duplicate-imports` rule was removed after review because it was primarily stylistic.

## Validation

- Oxlint: passed; no errors from the nine enabled rules. Existing warnings are from rules outside this batch.
- Oxfmt: passed for all 289 files.
- TypeScript: passed with `tsconfig.typecheck.json`.
- Jest: 17 suites and 110 tests passed when excluding the existing API integration suite that requires a server on `127.0.0.1:5000`.
- Next production build: passed.
- `git diff --check`: passed.

## Notes

The repository's actual default branch is `bundlephobia`; there is no remote `main` branch. This PR targets `bundlephobia`.


scripts/cleanup-old-keys.ts
```diff
@@ -172,27 +172,21 @@ async function processBackupFile(
...
           packagesRemoved++
-          action = 'Pruned'
-          if (!searchInfo) {
-            reason = 'Not found in searches-v2'
-          } else if (searchInfo.count <= 1) {
-            reason = `Search count (${searchInfo.count}) <= 1`
-          } else if (searchInfo.lastSearched < sixMonthsAgo) {
-            reason = 'Last searched more than 6 months ago'
-          }
+          const reason = !searchInfo
```
pastelsky: if-else was cleaner no?

cache-service/middlewares/exports-size.middleware.ts
```diff
@@ -6,12 +6,12 @@ import firebase from 'firebase'
...
-  CacheEntry,
-  CacheKey,
-  CacheRequest,
-  CacheRequestBody,
+import {
+  readCacheSnapshot,
+  type CacheEntry,
+  type CacheKey,
+  type CacheRequest,
+  type CacheRequestBody,
```
pastelsky: This rule is stylistic and not the most useful.

pastelsky (commented):
Done

PR #1110, opened 2026-09-13
Enable second anti-slop lint batch

pastelsky:
## What this changes

Enables the second focused anti-slop lint batch as errors in Oxlint and fixes all existing violations on the current default branch.

Rules enabled:

- `no-unused-vars`
- `no-shadow`
- `eqeqeq`
- `no-negated-condition`
- `preserve-caught-error`

The fixes remove dead bindings, make variable scopes unambiguous, replace loose equality with explicit null/undefined checks where that was the intended behavior, preserve caught errors as causes, and simplify negated conditional branches without changing behavior.

Test-package fixtures are excluded from this production lint batch; they intentionally model invalid or incomplete packages.

## Validation

- `oxlint .` — passed for all enabled rules; existing warnings from other rule families remain.
- `oxfmt --check .` — passed.
- `tsc --project tsconfig.typecheck.json --noEmit` — passed.
- Jest — 17 suites / 110 tests passed.
- `next build` — passed.
- `git diff --check` — passed.

The repository's actual default branch is `bundlephobia`; there is no `main` branch, so this PR targets the latest `bundlephobia` commit.


test-packages/missing-dependency-error/index.js
```diff
@@ -1,2 +1,3 @@
 const test = require('missing-package')
+void test
```
pastelsky: Test packages need not apply these rules.

pastelsky (commented):
Check

